PR #1, opened 2018-05-04
Create Prometheus using ECS

idavidmcdonald:
### This PR
Initial PR for ticket https://trello.com/c/j7p86jZo/382-mvp-prometheus-running-in-ecs

Uses terraform to create an AWS ECS node running prometheus in one availability zone accessible by load balancer URL. Prometheus is accessed directly by the ALB and there is currently no nginx in front of it.

### TODO
- Add tags to ECS services resources

### Things not included in this PR
- Automated terraform fmt or CI tests to confirm terraform fmt has been run
- Automate terraform-docs generation or CI tests to confirm the latest docs have been generated
- Scripts to make it easier to spin all the projects up and down in one command (same for generating docs for all the projects)


README.md
```diff
@@ -0,0 +1,122 @@
...
+After this change, your credentials should be only asked the first time you use the tool after start/restart the machine.
+
+## ECS
+
+### Newest ECS AMI
+
+To see the latest ECS Optimized Amazon Linux AMI information in your
+default region, run this AWS CLI command:
+
+    aws ssm get-parameters --names /aws/service/ecs/optimized-ami/amazon-linux/recommended
```
MuriloDalRi: @deanwilson we copied this across from your spike repo, when is this supposed to be used?

deanwilson: This would be used when moving to an updated ECS AMI, it stops you having to click through the AWS website docs for an hour.

https://github.com/deanwilson/ecs-monitoring/blob/master/terraform/projects/app-ecs-nodes/main.tf#L23

You can do some lookup magic to make EC2 pick the newest AMI but I intensely dislike that as you don't know what your base AMI is.

terraform/projects/app-ecs-albs/main.tf
```diff
@@ -0,0 +1,145 @@
+/**
+* ## Project: app-ecs-albs
+*
+* Create ALBs for the ECS cluster
+*
+* This may be merged back in to the nodes project
```
MuriloDalRi: @deanwilson does this comment still stand?

deanwilson: Probably not anymore.

terraform/projects/app-ecs-albs/main.tf
```diff
@@ -0,0 +1,145 @@
...
+
+resource "aws_lb" "monitoring_external_alb" {
+  name               = "${var.stack_name}-ext-alb"
+  internal           = false
+  load_balancer_type = "application"
+  security_groups    = ["${data.terraform_remote_state.infra_security_groups.monitoring_external_sg_id}"]
+
+  subnets = [
+    "${element(data.terraform_remote_state.infra_networking.public_subnets, 1)}",
+    "${element(data.terraform_remote_state.infra_networking.public_subnets, 2)}",
```
MuriloDalRi: Is this the right number of subnets?

terraform/projects/app-ecs-albs/main.tf
```diff
@@ -0,0 +1,145 @@
...
+  /* access_logs { */
+  /*   bucket  = "${aws_s3_bucket.lb_logs.bucket}" */
+  /*   prefix  = "test-lb" */
+  /*   enabled = true */
+  /* } */
+
+  tags = "${merge(
+    local.default_tags,
+    var.additional_tags,
+    map("Stackname", "${var.stack_name}"),
```
MuriloDalRi: Do we still want the stack name

terraform/projects/app-ecs-albs/main.tf
```diff
@@ -0,0 +1,145 @@
...
+    map("Name", "${var.stack_name}-ecs-monitoring")
```
MuriloDalRi: Same as above

terraform/projects/app-ecs-albs/main.tf
```diff
@@ -0,0 +1,145 @@
...
+
+resource "aws_lb_target_group" "monitoring_external_tg" {
+  name     = "${var.stack_name}-ext-tg"
+  port     = 80
+  protocol = "HTTP"
+  vpc_id   = "${data.terraform_remote_state.infra_networking.vpc_id}"
+
+  health_check {
+    interval            = "10"
+    path                = "/graph"
```
MuriloDalRi: We should explain why we're using this path

terraform/projects/app-ecs-nodes/main.tf
```diff
@@ -0,0 +1,185 @@
...
+variable "ecs_image_id" {
+  type        = "string"
+  description = "AMI ID to use for the ECS nodes"
+  default     = "ami-2d386654"
+}
+
+variable "ecs_instance_type" {
+  type        = "string"
+  description = "ECS Node instance type"
+  default     = "t2.medium"
```
MuriloDalRi: What should this be?

terraform/projects/app-ecs-albs/main.tf
deanwilson: To start with you probably only need one but it should be in multiple AZs, I'd say 3, by the end of the beta.

terraform/projects/app-ecs-albs/main.tf
deanwilson: If you want to run two versions of the environment at once, for example a load testing and an upgrade test, how would you determine which is which if you remove the stackname? Not saying to keep it, but ensure you have some way to tell.

terraform/projects/app-ecs-nodes/main.tf
deanwilson: The instance type will come out of the capacity planning trello story. If you're going to point this at DGU then you'll want at least as big as you currently have.

README.md
idavidmcdonald: Thanks, have added a small sentence to clarify when to use this.

terraform/projects/app-ecs-albs/main.tf
idavidmcdonald: Thanks, will go 1 AZ for the moment then.

terraform/projects/app-ecs-nodes/main.tf
idavidmcdonald: Cool, will bump up 👍

terraform/projects/app-ecs-nodes/main.tf
```diff
@@ -0,0 +1,179 @@
...
+  config {
+    bucket = "${var.remote_state_bucket}"
+    key    = "infra-security-groups.tfstate"
+    region = "${var.aws_region}"
+  }
+}
+
+## Resources
+
+resource "null_resource" "node_autoscaling_group_tags" {
```
idavidmcdonald: Could someone please clarify why we need to do this? We appear to be doing tags in a different way for `monitoring_external_alb.tags` so trying to understand if there is intentional difference in behaviour?

terraform/projects/app-ecs-albs/main.tf
idavidmcdonald: Sounds good, we will keep

README.md
```diff
@@ -0,0 +1,138 @@
...
+
+Every time you execute AWS vault it may ask for the credentials to access the keychain.
+
+To avoid this, we can follow the next steps (in OS X):
+
+1. Open the 'Keychain Access' utility.
+2. In the menu select "File > Add Keychain..."
+3. Select the "aws-vault.keychain". It can be found in `~/Library/Keychains/`.
+4. Once it is added, right click in it (it shows in the left hand side under 'Keychains') and select `Change Settings for Keychain aws-vault`.
+5. Uncheck `Lock after X minutes of inactivity` and `Look when sleeping`
```
philandstuff: I got a bit worried when reading this.  If I uncheck these options, when will my keychain ever lock again?

(I don't know a lot about OS X Keychain)

philandstuff: after a bit of digging I think this advice is wrong.

There's two different things going on:
 - access for aws-vault app to the keychain
 - whether the keychain is locked or unlocked

By default, you have to type the password each time you want the app to access the keychain, whether the keychain itself is locked or not.  To prevent this, you click "always allow" instead of "allow" in the pop up dialog.

Also by default, the keychain locks after your computer sleeps, or after 5 minutes of inactivity.  After that, any use of the keychain requires the password again.

I think we should recommend users click "always allow" to allow aws-vault to access the keychain without prompting every time, but do not change the "lock after X minutes" or "lock when sleeping" settings.

README.md
```diff
@@ -0,0 +1,138 @@
...
+stack. Inside the `stacks` directory you will find a pair of files
+for each stack, a `.backend` and a `.tfvars`. Make a copy of an existing
+pair and change the values to suit your new name. The `bucket`
+and `remote_state_bucket` settings in these files must match the bucket you
+created above.
+
+You should also ensure that the AWS account you are creating your environment in has an SSH key pair
+set up called `ecs-monitoring-ssh-test`. You should do this manually using the AWS web console. You
+will need to download the private key for this key pair when you create it if you wish to SSH in to
+the ECS node.
```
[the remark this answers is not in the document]
idavidmcdonald: Yeah, might be fair to remove it for now so there is no key pair for the instance.

We previously SSH'd in to the box but required us manually setting up a security group and ELB to go through

PR #2, opened 2018-05-10
Rename `nodes` to `instances` to follow AWS naming conventions

JonathanHallam:
ECS describes the EC2 instances that run containers inside them as `Container instances` rather than `nodes` so we will follow this convention for readability.


terraform/projects/app-ecs-instances/main.tf
```diff
@@ -133,23 +133,23 @@ resource "aws_ecs_cluster" "prometheus_cluster" {
...
     cluster_name = "${local.cluster_name}"
   }
 }
 
-module "ecs_node" {
+module "ecs_instance" {
   source = "terraform-aws-modules/autoscaling/aws"
 
-  name = "${var.stack_name}-ecs-node"
+  name = "${var.stack_name}-ecs_instance"
```
philandstuff: can this be `kebab-case` not mixed `snake_and_kebab-case`?

terraform/projects/app-ecs-instances/main.tf
```diff
@@ -133,23 +133,23 @@ resource "aws_ecs_cluster" "prometheus_cluster" {
...
 
   key_name = "${var.ecs_instance_ssh_keyname}"
 
   # Launch configuration
-  lc_name = "${var.stack_name}-ecs-node"
+  lc_name = "${var.stack_name}-ecs_instance"
```
philandstuff: ditto

terraform/projects/app-ecs-instances/main.tf
```diff
@@ -178,13 +178,13 @@ module "ecs_node" {
     local.default_tags,
     var.additional_tags,
     map("Stackname", "${var.stack_name}"),
-    map("Name", "${var.stack_name}-ecs_node")
+    map("Name", "${var.stack_name}-ecs_instance")
```
philandstuff: ditto (even if the old code was mixed)

idavidmcdonald:
Thanks Phil, new commit added to address this

PR #3, opened 2018-05-11
capture logs from ECS tasks into CloudWatch

philandstuff:
this creates a cloudwatch logs group for the stack, and ships logs
from each of our ECS tasks into CloudWatch.  This means that when a
task fails, we can get the output from the command and see what
exactly went wrong!

### DIFF
--- a/terraform/projects/app-ecs-services/main.tf
+++ b/terraform/projects/app-ecs-services/main.tf
@@ -82,9 +82,6 @@ data "terraform_remote_state" "app_ecs_albs" {
 resource "aws_cloudwatch_log_group" "task_logs" {
   name = "${var.stack_name}"
   retention_in_days = 7
-  tags = {
-    Environment = "${var.stack_name}"
-  }
 }
 
 ## Outputs
--- a/terraform/projects/app-ecs-services/main.tf
+++ b/terraform/projects/app-ecs-services/main.tf
@@ -41,6 +41,10 @@ provider "aws" {
   region  = "${var.aws_region}"
 }
 
+provider "template" {
+  version = "~> 1.0.0"
+}
+
 ## Data sources
 
 data "terraform_remote_state" "infra_networking" {
@@ -75,6 +79,13 @@ data "terraform_remote_state" "app_ecs_albs" {
 
 ## Resources
 
+resource "aws_cloudwatch_log_group" "task_logs" {
+  name = "${var.stack_name}"
+  retention_in_days = 7
+  tags = {
+    Environment = "${var.stack_name}"
+  }
+}
 
 ## Outputs
 
--- a/terraform/projects/app-ecs-services/task-definitions/git-puller.json
+++ b/terraform/projects/app-ecs-services/task-definitions/git-puller.json
@@ -16,7 +16,8 @@
       "logDriver": "awslogs",
       "options": {
         "awslogs-group": "${log_group}",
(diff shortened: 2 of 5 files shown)


terraform/projects/app-ecs-services/main.tf
```diff
@@ -75,6 +79,13 @@ data "terraform_remote_state" "app_ecs_albs" {
 
 ## Resources
 
+resource "aws_cloudwatch_log_group" "task_logs" {
+  name = "${var.stack_name}"
+  retention_in_days = 7
+  tags = {
+    Environment = "${var.stack_name}"
```
idavidmcdonald: These tags could be more consistent with the rest of our tags in the other projects please (e.g. adding `Stackname`, using additional vars as well)

terraform/projects/app-ecs-services/task-definitions/git-puller.json
```diff
@@ -11,6 +11,14 @@
...
       }
     ],
-    "command": ["clone", "https://github.com/alphagov/prometheus-config.git", "/configs"]
+    "command": ["clone", "https://github.com/alphagov/prometheus-config.git", "/configs"],
+    "logConfiguration": {
+      "logDriver": "awslogs",
+      "options": {
+        "awslogs-group": "${log_group}",
+        "awslogs-region": "${region}",
+        "awslogs-stream-prefix": "git-puller"
```
idavidmcdonald: Do we need the stream prefixes? They seem kind of redundant as it is a duplicated in the log stream name

terraform/projects/app-ecs-services/task-definitions/prometheus-server.json
```diff
@@ -20,6 +20,14 @@
...
         "containerPath": "/etc/prometheus/alerts/alerts.default"
       }
-    ]
+    ],
+    "logConfiguration": {
+      "logDriver": "awslogs",
+      "options": {
+        "awslogs-group": "${log_group}",
+        "awslogs-region": "${region}",
+        "awslogs-stream-prefix": "prometheus"
```
idavidmcdonald: Ditto to above

terraform/projects/app-ecs-services/main.tf
philandstuff: can you be more specific? I can't find any other examples of us setting Environment in this repo

terraform/projects/app-ecs-services/task-definitions/git-puller.json
philandstuff: I tried without and it turns out that specifying a stream prefix has nontrivial extra behaviour which we want. see 148fc06 for details.

terraform/projects/app-ecs-services/main.tf
idavidmcdonald: https://github.com/alphagov/prometheus-aws-configuration-beta/blob/master/stacks/staging.tfvars Is where we set the environment as an additional tag and then we set all the tags on resources like https://github.com/alphagov/prometheus-aws-configuration-beta/blob/master/terraform/projects/infra-security-groups/main.tf#L85

terraform/projects/app-ecs-services/task-definitions/git-puller.json
idavidmcdonald: Thanks for investigating 👍

PR #8, opened 2018-05-16
Persistent storage

JonathanHallam:
https://trello.com/c/ak0GUE0i/386-add-persistent-storage-for-prometheus

This adds the necessary terraform to initialise and format an EBS volume to be used for the persistent storage for our Prometheis. This means that in the event that an EC2 instance goes down we will retain metrics that have been gathered on it.

### DIFF
--- a/terraform/projects/app-ecs-instances/ecs-iam-policy.tf
+++ b/terraform/projects/app-ecs-instances/ecs-iam-policy.tf
@@ -31,6 +31,12 @@ data "aws_iam_policy_document" "ecs_instance_document" {
     resources = ["*"]
 
     actions = [
+      "ec2:DescribeNetworkInterfaces",
+      "ec2:AttachNetworkInterface",
+      "ec2:AttachVolume",
+      "ec2:DetachVolume",
+      "ec2:DescribeVolumeStatus",
+      "ec2:DescribeVolumes",
       "ecs:CreateCluster",
       "ecs:DeregisterContainerInstance",
       "ecs:DiscoverPollEndpoint",
--- a/terraform/projects/app-ecs-instances/ecs-iam-policy.tf
+++ b/terraform/projects/app-ecs-instances/ecs-iam-policy.tf
@@ -48,8 +48,14 @@ data "aws_iam_policy_document" "ecs_instance_document" {
       "logs:createlogstream",
       "logs:putlogevents",
     ]
+  }
 
-    resources = ["arn:aws:ec2:${var.aws_region}:${var.account_id}:volume/${aws_ebs_volume.prometheus_ebs_volume.id}"]
+  statement {
+    resources = [
+      "arn:aws:ec2:${var.aws_region}:${var.account_id}:volume/${aws_ebs_volume.prometheus_ebs_volume.id}",
+      "arn:aws:ec2:${var.aws_region}:${var.account_id}:instance/*"
+    ]
+
 
     actions = [
       "ec2:AttachVolume",
--- a/terraform/projects/app-ecs-instances/ecs-iam-policy.tf
+++ b/terraform/projects/app-ecs-instances/ecs-iam-policy.tf
@@ -26,36 +26,36 @@ EOF
 
 data "aws_iam_policy_document" "ecs_instance_document" {
   statement {
-    sid       = "ECSInstancePolicy"
+    sid = "ECSInstancePolicy"
+
     resources = ["*"]
 
     actions = [
-      "ec2:describenetworkinterfaces",
-      "ec2:attachnetworkinterface",
-      "ec2:describevolumestatus",
-      "ec2:describevolumes",
-      "ecs:createcluster",
-      "ecs:deregistercontainerinstance",
-      "ecs:discoverpollendpoint",
-      "ecs:poll",
-      "ecs:registercontainerinstance",
-      "ecs:starttelemetrysession",
-      "ecs:submit*",
-      "ecr:getauthorizationtoken",
-      "ecr:batchchecklayeravailability",
-      "ecr:getdownloadurlforlayer",
-      "ecr:batchgetimage",
-      "logs:createlogstream",
-      "logs:putlogevents",
+      "ec2:DescribeNetworkInterfaces",
+      "ec2:AttachNetworkInterface",
+      "ec2:DescribeVolumeStatus",
+      "ec2:DescribeVolumes",
+      "ecs:CreateCluster",
+      "ecs:DeregisterContainerInstance",
+      "ecs:DiscoverPollEndpoint",
+      "ecs:Poll",
+      "ecs:RegisterContainerInstance",
+      "ecs:StartTelemetrySession",
+      "ecs:Submit*",
+      "ecr:GetAuthorizationToken",
+      "ecr:BatchCheckLayerAvailability",
+      "ecr:GetDownloadUrlForLayer",
+      "ecr:BatchGetImage",
+      "logs:CreateLogStream",
+      "logs:PutLogEvents",
     ]
   }
 
   statement {
     resources = [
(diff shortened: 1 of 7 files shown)


terraform/projects/app-ecs-instances/ecs-iam-policy.tf
```diff
@@ -31,6 +31,12 @@ data "aws_iam_policy_document" "ecs_instance_document" {
     resources = ["*"]
 
     actions = [
+      "ec2:DescribeNetworkInterfaces",
+      "ec2:AttachNetworkInterface",
+      "ec2:AttachVolume",
```
philandstuff: does this allow the instance to attach *any* volume?  is it worth trying to be a bit more restrictive?

terraform/projects/app-ecs-instances/instance-user-data.tpl
```diff
@@ -1,8 +1,44 @@
...
+  if ! [[ -e /dev/$DEVICE ]] ; then
+    sleep 1
+  else
+    break
+  fi
+  x=$((x+1))
+done
+
+# Format and mount volume
+if sudo file -s /dev/$DEVICE | grep -q "/dev/$DEVICE: data"; then
```
philandstuff: do we need `sudo` in user data? I thought it ran as root anyway?

terraform/projects/app-ecs-instances/instance-user-data.tpl
```diff
@@ -1,8 +1,44 @@
...
+  echo "[$(date '+%H:%M:%S %d-%m-%Y')] attach-volume: /dev/$DEVICE does not contain any partition, beginning to format disk"
+  sudo mkfs -t ext4 /dev/$DEVICE
+else
+  echo "[$(date '+%H:%M:%S %d-%m-%Y')] attach-volume: /dev/$DEVICE is already formatted: $(file -s /dev/$DEVICE)"
+fi
+
+# Allow prometheus container user access to read/write/execute within container
+sudo mkdir -p /ecs/prometheus_data
+sudo mount /dev/$DEVICE /ecs/prometheus_data
+sudo chmod 777 /ecs/prometheus_data
```
philandstuff: is `chmod 777` really necessary? 😨 

(answer: no it's not. let's see if we can tighten this up)

terraform/projects/app-ecs-instances/main.tf
```diff
@@ -182,6 +184,23 @@ module "ecs_instance" {
   )}"
 }
 
+resource "aws_ebs_volume" "prometheus_ebs_volume" {
+  availability_zone = "${element(data.terraform_remote_state.infra_networking.az_names, 0)}"
+  size              = 500
+  type              = "gp2"
+
+  lifecycle {
+    prevent_destroy = true
```
philandstuff: 👍

terraform/projects/app-ecs-services/task-definitions/prometheus-server.json
```diff
@@ -3,7 +3,7 @@
     "name": "prometheus",
     "image": "prom/prometheus",
     "cpu": 2048,
-    "memory": 8192,
+    "memoryReservation": 8192,
```
philandstuff: I've opened #9 to deal with this - we can leave it for now but it'll get changed again

ejrowley:
Hi, has there been a discussion around using LVM, LVM makes online volume changes a little less risky IMHO.

terraform/projects/app-ecs-services/task-definitions/prometheus-server.json
```diff
@@ -19,6 +19,10 @@
       {
         "sourceVolume": "alert-config",
         "containerPath": "/etc/prometheus/alerts"
+      },
+      {
+        "sourceVolume": "prometheus-timeseries-storage",
+        "containerPath": "/prometheus"
```
philandstuff: can we document that this is default directory that our docker image expects? i was confused for a while because i couldn't see how this got wired up

philandstuff:
@ejrowley good point, lets discuss tomorrow

terraform/projects/app-ecs-instances/ecs-iam-policy.tf
MuriloDalRi: I think we could probably be more specific: https://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_AttachVolume.html

MuriloDalRi: Actually it's done in here: https://github.com/alphagov/prometheus-aws-configuration-beta/pull/8/files?utf8=%E2%9C%93&diff=unified&w=1#diff-b9fa37e9b366d987de5582400d4a8326R8

philandstuff: that line doesn't restrict permissions though.  The instance could still, if compromised somehow, attach arbitrary volumes.  I just wondered if we could specify `resources` to only allow attaching the prometheus volume.

terraform/projects/app-ecs-instances/instance-user-data.tpl
JonathanHallam: we didn't, have changed

terraform/projects/app-ecs-instances/instance-user-data.tpl
JonathanHallam: I tried running at 755, it wouldn't go through. Ed recommended that we wait until we can have access to the box to change groups around and work out what we can get away with. I've left it as is for now

terraform/projects/app-ecs-services/task-definitions/prometheus-server.json
JonathanHallam: 👍

terraform/projects/app-ecs-instances/ecs-iam-policy.tf
```diff
@@ -31,6 +31,10 @@ data "aws_iam_policy_document" "ecs_instance_document" {
     resources = ["*"]
 
     actions = [
+      "ec2:DescribeNetworkInterfaces",
```
idavidmcdonald: Do we need all of these new permissions?

terraform/projects/app-ecs-instances/main.tf
```diff
@@ -5,6 +5,12 @@
 *
 */
 
+variable "account_id" {
+  type        = "string"
+  description = "Our AWS account ID"
+  default     = "0"
```
idavidmcdonald: Could we document in the readme that people should be setting the account idea in their tfvars file so we can spin this up please? Actually, passing it in to the command line is probably better given it removes the risk that people commit them in the staging and production tfvars.

I kind of dislike the fact that we manually need to know the account ID to run any of the terraform. Might be worth double checking if there is any way we could get around this though...

idavidmcdonald: Maybe the terraform-pass integration so we could get it from reng-pass could be an option...

philandstuff:
this PR now has conflicts - can you resolve please

terraform/projects/app-ecs-instances/main.tf
philandstuff: did some spiking and it looks like we can use `data "aws_caller_identity"`: https://www.terraform.io/docs/providers/aws/d/caller_identity.html

terraform/projects/app-ecs-services/prometheus-service.tf
```diff
@@ -89,6 +89,17 @@ resource "aws_ecs_task_definition" "prometheus_server" {
...
+  volume {
+    name      = "prometheus-timeseries-storage"
+    host_path = "/ecs/prometheus_data"
+  }
+
+  # We mount this at /prometheus which is the expected location for the prom/prometheus docker image
+  volume {
+    name      = "prometheus-timeseries-storage"
+    host_path = "/ecs/prometheus_data"
+  }
```
philandstuff: oops

PR #10, opened 2018-05-17
WIP - travis

idavidmcdonald:


.travis.yml
```diff
@@ -0,0 +1,10 @@
+env:
+  - TERRAFORM_VERSION=0.11.7 TERRAFORM_FILE_NAME=terraform_${TERRAFORM_VERSION}_linux_amd64.zip TERRAFORM_DOWNLOAD_URL=https://releases.hashicorp.com/terraform/${TERRAFORM_VERSION}/${TERRAFORM_FILE_NAME}
+install:
+  - wget ${TERRAFORM_DOWNLOAD_URL}
```
philandstuff: do we want to cache this somehow, so we don't have to redownload terraform on every CI run? https://docs.travis-ci.com/user/caching/

idavidmcdonald: So I think Terraform is distributed as a binary package (https://www.terraform.io/intro/getting-started/install.html). The Travis docs say it is not particularly useful to cache a compiled binary (https://docs.travis-ci.com/user/caching/#Things-not-to-cache).

If my understanding is correct (I'm not that familiar with binary packages), it means this wouldn't be particularly needed. If you could just double check my logic that would be great thanks.

PR #14, opened 2018-05-22
Add setup.sh and Makefile

kentsanggds:
## What

In order to make it easier to develop, a `setup.sh` and `Makefile` that uses the `setup.sh` file have been added.

The README has been updated to explain how to use them and a `environment_sample.sh` file has been added as the `setup.sh` file needs env vars to work.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,17 +1,12 @@
 .DEFAULT_GOAL := help
-SHELL := /bin/zsh
-DATE = $(shell date +%Y-%m-%d:%H:%M:%S)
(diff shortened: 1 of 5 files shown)


Makefile
```diff
@@ -0,0 +1,43 @@
+.DEFAULT_GOAL := help
+SHELL := /bin/zsh
```
MuriloDalRi: Is this going to work on zsh only?

kentsanggds: Not sure, why not try it

PR #15, opened 2018-05-23
Get targets from s3 

kentsanggds:
## What 

Gets the targets from the S3 bucket `gds-prometheus-targets`

Currently the way it grabs the targets is defining the service as a task definition which is set to be always running. This means that it will run and then complete and will re-run, we may want to revisit this for a better solution involving triggers from s3.

Also we thought that the default Nginx resolver would work to resolve the targets, but we had to introduce a resolver in the `paas-proxy.conf` and set it to the Google resolver address `8.8.8.8`. So there may be work around getting it to resolve to the default resolver and working with AWS address `10.0.0.2`.


terraform/projects/app-ecs-services/prometheus-service.tf
```diff
@@ -120,6 +134,34 @@ resource "aws_ecs_service" "prometheus_server" {
   }
 }
 
+resource "aws_ecs_service" "targets_grabber" {
+  name            = "${var.stack_name}-targets_grabber"
```
philandstuff: can we not mix camel and snake case please? all hyphens would be better

terraform/projects/app-ecs-services/task-definitions/targets_grabber.json
```diff
@@ -0,0 +1,26 @@
...
+    "logConfiguration": {
+      "logDriver": "awslogs",
+      "options": {
+        "awslogs-group": "${log_group}",
+        "awslogs-region": "${region}",
+        "awslogs-stream-prefix": "s3-targets"
+      }
+    }
+  }
+]
```
philandstuff: can you set your editor to always put a newline at end of file?

PR #18, opened 2018-05-24
Add a Jumpbox to dev stacks

ejrowley:
An ssh key is created based on the users public key (it assumes the
users public key is available as `~/.ssh/id_rsa.pub`), so at the moment connecting to the instance is only available for the user who is applying the jumpbox project.

In order to get the jumpbox to work you will need to ensure that you have `jq` installed and you allow your ssh key to be forwarded, details of which can be found in the README.

It's be a good idea to start with a fresh stack in order to ensure that your ssh key is being forwarded to the EC2 instance, so if you have an existing stack to destroy it and then apply the terraform to build it again.

To connect to the EC2 instance, rather than jumping onto the jumpbox and then ssh-ing on the the instance the `make jump` task will do both steps for you.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -69,6 +69,10 @@ destroy-list: ## Destroy terraform for a list of projects, make destroy-list lis
 taint: ## Taint a resource, make taint project=<project name> resource=<resource name>
 	. ./setup.sh -t ${project} ${resource}
 
+.PHONY: jump
+jump: ## Jump onto the first instance on a dev environment
+	. ./setup.sh -j
+
 .PHONY: docs
 docs: ## Update all terraform docs
 	. ./tools/update-docs.sh
--- a/setup.sh
+++ b/setup.sh
@@ -1,4 +1,6 @@
 #!/bin/bash
+set -e
+
 DEFAULT_DEV_TARGETS_S3_BUCKET=gds-prometheus-targets-dev
 TERRAFORM_BUCKET=${TERRAFORM_BUCKET}
 TERRAFORMPATH=$(which terraform)
@@ -8,8 +10,8 @@ ROOTPROJ=$(pwd)
 TERRAFORMPROJ=$(pwd)/terraform/projects/
 SHARED_DEV_SUBDOMAIN_RESOURCE=aws_route53_zone.shared_dev_subdomain
 SHARED_DEV_DNS_ZONE=Z3702PZTSCDWPA  # this is the dev.gds-reliability.engineering DNS hosted zone ID
-declare -a COMPONENTS=("infra-networking" "infra-security-groups" "app-ecs-instances" "app-ecs-albs" "app-ecs-services")
-declare -a COMPONENTSDESTROY=("app-ecs-services" "app-ecs-albs" "app-ecs-instances" "infra-security-groups" "infra-networking")
+declare -a COMPONENTS=("infra-networking" "infra-security-groups" "infra-jump-instance" "app-ecs-instances" "app-ecs-albs" "app-ecs-services")
+declare -a COMPONENTSDESTROY=("app-ecs-services" "app-ecs-albs" "app-ecs-instances" "infra-jump-instance" "infra-security-groups" "infra-networking")
 
 
 ############ Actions #################
@@ -32,10 +34,11 @@ if [ -e "${ROOTPROJ}/stacks/${ENV}.tfvars" ] ; then
         echo "${ENV}.tfvars exists"
 else
 cat <<EOF >stacks/${ENV}.tfvars
+dev_environment = "true"
+ecs_instance_ssh_keyname = "${ENV}-jumpbox-key"
+prometheus_subdomain = "${ENV}"
 remote_state_bucket = "${TERRAFORM_BUCKET}"
 stack_name = "${ENV}"
-dev_environment = "true"
-prometheus_subdomain = "${ENV}"
 targets_s3_bucket="$DEFAULT_DEV_TARGETS_S3_BUCKET"
 prometheis_total=1
 additional_tags = {
@@ -98,6 +101,11 @@ remove_shared_dev_route53 () {
 
 init () {
 # Init a terraform project
+        # Only init the jump box for dev stacks
+        if [ $DEV_ENVIRONMENT != 'true' -a "$1" = 'infra-jump-instance' ] ; then
+                return
+        fi
+
         echo $1
 
         cd $TERRAFORMPROJ$1
@@ -114,6 +122,11 @@ plan () {
 
 apply () {
 # Apply a terraform project
+        # Only create the jump box for dev stacks
+        if [ $DEV_ENVIRONMENT != 'true' -a "$1" = 'infra-jump-instance' ] ; then
+                return
+        fi
+
         echo $1
 
         cd $TERRAFORMPROJ$1
@@ -129,6 +142,10 @@ apply () {
 
 destroy () {
 # Destroy a terraform project
+        # Only destroy the jump box for dev stacks
+        if [ $DEV_ENVIRONMENT != 'true' -a "$1" = 'infra-jump-instance' ] ; then
+                return
+        fi
         echo $1
 
         cd $TERRAFORMPROJ$1
@@ -151,6 +168,33 @@ taint() {
 
         cd $TERRAFORMPROJ$1
         aws-vault exec ${PROFILE_NAME} -- $TERRAFORMPATH taint $2
+}
+
+jumpbox() {
+        if [ $DEV_ENVIRONMENT != 'true' ] ; then
+                echo "Jumpbox is only available for dev environments"
+                return
+        fi
+
+        EC2_DATA=$(aws-vault exec $PROFILE_NAME -- aws ec2 describe-instances --filters "Name=tag:Environment,Values=$ENV" "Name=instance-state-name,Values=running")
+
+        EC2_DATA_STR=$(echo "$EC2_DATA" | jq -c .) 
+        
+        if [ $EC2_DATA_STR = '{"Reservations":[]}' ] ; then
+                echo "No EC2 instances running"
+                return
+        fi
+
+        INSTANCE_IP=$(echo $EC2_DATA | jq -cr '[.Reservations[] | select(.Instances[]).Instances | first.PrivateIpAddress] | first')
+
+        JUMPBOX=jump.$ENV.dev.gds-reliability.engineering
+
+        # jumpbox fingerprint removed so that the jumpbox IP can change without affecting the connection to the instance
+        echo Remove $JUMPBOX fingerprint from ~/.ssh/known_hosts
+        ssh-keygen -R $JUMPBOX
+
+        echo "Connecting to instance: $INSTANCE_IP via jumpbox: ec2-user@$JUMPBOX"
+        ssh -At -oStrictHostKeyChecking=no ec2-user@$JUMPBOX ssh -oStrictHostKeyChecking=no ec2-user@$INSTANCE_IP
 }
 
 #################################
@@ -289,6 +333,9 @@ else
         ;;
         -t) echo "Taint a terraform resource: ${ENV}"
                 taint $2 $3
+        ;;
+        -j) echo "Jump onto instance: ${ENV}"
+                jumpbox
         ;;
         *) echo "Invalid option"
         ;;
--- a/terraform/modules/common/ami/main.tf
+++ b/terraform/modules/common/ami/main.tf
@@ -1,0 +1,29 @@
+## Variables
+
+variable "amazon_release" {
+  description = "The amazon release used to generate the encrypted ami"
+}
+
+## Data sources
+
+data "aws_ami" "source" {
+  most_recent = true
+
+  filter {
+    name   = "name"
+    values = ["amzn-ami-${var.amazon_release}-amazon-ecs-optimized"]
+  }
+
+  filter {
+    name   = "architecture"
+    values = ["x86_64"]
+  }
+
+  owners = ["amazon"]
+}
+
+## Outputs
+
+output "ami_id" {
+  value = "${data.aws_ami.source.id}"
+}
--- a/terraform/modules/jumpbox/main.tf
+++ b/terraform/modules/jumpbox/main.tf
@@ -1,0 +1,125 @@
+## Variables
+
+variable "subnet_id" {
+  description = "The subnet which the jumpbox will be added to"
+}
+
+variable "vpc_id" {
+  description = "The VPC ID to use on this service"
+}
+
+variable "ecs_optimised_ami_version" {
+  default = "2018.03.a"
+}
+
+variable "stack_name" {
+  description = "The name of the stack the jumpbox belongs to"
+}
+
+variable "aws_region" {
+  description = "The region which the jump box will be built within"
+}
+
+variable "security_groups" {
+  type        = "list"
+  description = "A list of security groups to extend"
+  default     = []
+}
+
+variable "allowed_cidrs" {
+  type        = "list"
+  description = "List of CIDRs which are able to access the jumpbox, default are GDS ips"
+
+  default = [
+    "213.86.153.212/32",
+    "213.86.153.213/32",
+    "213.86.153.214/32",
+    "213.86.153.235/32",
+    "213.86.153.236/32",
+    "213.86.153.237/32",
+    "85.133.67.244/32",
+  ]
+}
+
+## Resources
+
+resource "aws_key_pair" "ssh_key" {
+  key_name   = "${var.stack_name}-jumpbox-key"
+  public_key = "${file("~/.ssh/id_rsa.pub")}"
+}
+
+module "ami" {
+  source         = "../common/ami"
+  amazon_release = "${var.ecs_optimised_ami_version}"
+}
+
+resource "aws_security_group" "bastion_security_group" {
+  name        = "${var.stack_name}-bastion-sg"
+  vpc_id      = "${var.vpc_id}"
+  description = "Controls access to & from the bastion"
+}
+
+resource "aws_security_group_rule" "allow_ssh" {
+  security_group_id = "${aws_security_group.bastion_security_group.id}"
+
+  type        = "ingress"
+  from_port   = 22
+  to_port     = 22
+  protocol    = "tcp"
+  cidr_blocks = ["${var.allowed_cidrs}"]
+}
+
+resource "aws_security_group_rule" "allow_all_egress" {
+  security_group_id = "${aws_security_group.bastion_security_group.id}"
+
+  type        = "egress"
+  from_port   = 0
+  to_port     = 0
+  protocol    = "-1"
+  cidr_blocks = ["${var.allowed_cidrs}"]
+}
+
+resource "aws_instance" "instance" {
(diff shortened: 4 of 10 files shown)


terraform/modules/common/ami/main.tf
```diff
@@ -0,0 +1,25 @@
+data "aws_ami" "source" {
+  most_recent = true
+
+  filter {
+    name   = "name"
+    values = ["ubuntu/images/hvm/ubuntu-${var.ubuntu_release}-*"]
```
philandstuff: our existing images are amazon linux - can we stick to that? (we went with it to have the baked-in optimized drivers for networking and i/o)

terraform/modules/common/ami/outputs.tf
```diff
@@ -0,0 +1,3 @@
+output "ami_id" {
+  value = "${data.aws_ami.source.id}"
+}
```
philandstuff: I don't think we need separate files for variables and outputs. @surminus explained why better than I could: https://surminus.github.io/post/terraform-structures-and-layouts/ (see section "variables and outputs")

terraform/modules/common/instance-profile/main.tf
```diff
@@ -0,0 +1,4 @@
+resource "aws_iam_instance_profile" "profile" {
+  name = "${var.role}-${var.product}-${var.environment}-instance-profile"
+  role = "${var.role}"
+}
```
philandstuff: i don't see where this module gets used. it also looks far too small to be something worth extracting into a module

terraform/modules/jumpbox/main.tf
```diff
@@ -0,0 +1,38 @@
...
+
+  type        = "ingress"
+  from_port   = 0
+  to_port     = 0
+  protocol    = "tcp"
+  cidr_blocks = ["${var.allowed_cidrs}"]
+}
+
+module "jump" {
+  source                      = "../common/instance"
```
philandstuff: i suspect our `terraform fmt` travis check is complaining about the alignment of all the `=` here

terraform/modules/jumpbox/main.tf
```diff
@@ -0,0 +1,38 @@
...
+  ami_id                      = "${module.ami.ami_id}"
+  ssh_key_name                = "${module.ssh_key.ssh_key_name}"
+  subnet_id                   = "${var.subnet_id}"
+  project                     = "${var.project}"
+  stack_name                  = "${var.stack_name}"
+  public_ip                   = true
```
philandstuff: this is probably fine for now, but i wonder if it's worth assigning an EIP?

philandstuff:
@ejrowley can you update this PR please?

DavidJeche:
I don't understand why we using a separate module for this when it should be something that should possibly be used in conjunction with the instances deployment. 

We could set a variable that if true the Jump box is deployed. This reduce's the amount of code & complication.

DavidJeche:
I am just wondering about the private hosted zone that we where going to have internally. The intent was to allow us to jump to the instances more easily using DNS when inside the VPC  ?

DavidJeche:
I think it is fine for the time being. I cannot seem to see the private DNS name attached to the instance, however this is not the most important things right. Private DNS names are luxury at the moment.

terraform/modules/common/ami/main.tf
```diff
@@ -0,0 +1,33 @@
+## Variables
+
+variable "amazon_release" {
+  description = "The amazon release used to generate the encrypted ami"
+}
+
+variable "stack_name" {
```
idavidmcdonald: Do we need the variable in this file given it's not used directly in it?

terraform/modules/common/instance/main.tf
```diff
@@ -0,0 +1,63 @@
...
+
+  key_name = "${var.ssh_key_name}"
+
+  associate_public_ip_address = "${var.public_ip}"
+
+  security_groups             = ["${var.security_groups}"]
+  associate_public_ip_address = "${var.public_ip}"
+
+  tags {
+    Name      = "jumpbox.${var.stack_name}.dmz"
```
idavidmcdonald: Should we have something jumpbox specific in the general instance module?

terraform/modules/common/instance/main.tf
```diff
@@ -0,0 +1,63 @@
...
+}
+
+variable "security_groups" {
+  description = "A list of security groups which will be added to the instance. If the list is empty the default securit group will be used"
+  type        = "list"
+}
+
+variable "public_ip" {
+  description = "When this is true a public ip is added to the instance"
+  default     = false
```
idavidmcdonald: See https://www.terraform.io/docs/configuration/variables.html#booleans as I believe it is suggested to use `"false"`

terraform/modules/common/instance/main.tf
```diff
@@ -0,0 +1,63 @@
...
+}
+
+## Resources
+
+resource "aws_instance" "instance" {
+  ami           = "${var.ami_id}"
+  subnet_id     = "${var.subnet_id}"
+  instance_type = "t2.small"
```
idavidmcdonald: Instance type could be put in a variable with t2.small as the default as it seems a bit specific to the jumpbox (I don't mind if you don't address this comment too much by the way)

terraform/modules/common/ssh_key/main.tf
```diff
@@ -0,0 +1,18 @@
+## Variables
+
+variable "stack_name" {
+  description = "The name of the stack within which the resource resides"
+}
+
+## Resources
+
+resource "aws_key_pair" "ssh_key" {
+  key_name   = "${var.stack_name}-jumpbox-key"
```
idavidmcdonald: Do we want to have `jumpbox` in here as I believe it is supposed to be generic? Potentially instead of taking `stack_name` as a variable to the module we could take `key_name` and let the caller be responsible for naming it?

terraform/modules/common/ami/main.tf
```diff
@@ -0,0 +1,33 @@
+## Variables
+
+variable "amazon_release" {
```
idavidmcdonald: Is this expected to always be a version? Would appending `_version` on the end help communicate this if so? Or maybe `ami_release_version`?

terraform/modules/common/ami/main.tf
```diff
@@ -0,0 +1,33 @@
...
+}
+
+## Data sources
+
+data "aws_ami" "source" {
+  most_recent = true
+
+  filter {
+    name   = "name"
+    values = ["amzn-ami-${var.amazon_release}-amazon-ecs-optimized"]
```
idavidmcdonald: It feels like this ami module is specific to ECS optimised instances so maybe the module should be called `ecs-ami` rather than just `ami`?

terraform/modules/jumpbox/main.tf
```diff
@@ -0,0 +1,105 @@
...
+  security_group_id = "${aws_security_group.bastion_security_group.id}"
+
+  type        = "egress"
+  from_port   = 0
+  to_port     = 0
+  protocol    = "-1"
+  cidr_blocks = ["${var.allowed_cidrs}"]
+}
+
+module "jump" {
```
idavidmcdonald: I wonder if a more appropriate name for this module would be `jumpbox_instance`?

terraform/projects/infra-jump-instance/main.tf
```diff
@@ -0,0 +1,109 @@
...
+variable "stack_name" {
+  type        = "string"
+  description = "Unique name for this collection of resources"
+  default     = "ecs-monitoring"
+}
+
+variable "enable_jumpbox" {
+  type        = "string"
+  description = "Boolean, when true a jumpbox will be deployed"
+  default     = false
```
idavidmcdonald: https://www.terraform.io/docs/configuration/variables.html#booleans

terraform/projects/infra-jump-instance/README.md
```diff
@@ -0,0 +1,56 @@
...
+When auto scaling groups are deployed into the private they are not accessible
+from public networks since they do not have a public IP address. In order to
+access them a jump box is in created from within this project. The instance has
+got a public IP address which can be used in order to access.
+
+There is a number of firewall rules that have been created to allow you access to all
+public + private subnets on any port. The instance is currently only allowed ssh access
+to port number 22 from any public network.
+
+In order to properly utilize this instance, you need to enable ssh-agent key forwarding.
```
idavidmcdonald: Wondering if all instructions to get this working should be in the README as it feels like someone might not know to look here?

terraform/modules/common/ssh_key/main.tf
```diff
@@ -0,0 +1,18 @@
+## Variables
+
+variable "stack_name" {
+  description = "The name of the stack within which the resource resides"
+}
+
+## Resources
+
+resource "aws_key_pair" "ssh_key" {
```
idavidmcdonald: Is there a particular advantage to having the ssh_key as it's own file? It feels like it's closely enough tied to the jumpbox (as in we have no other need for one) that it could go into that file in the same way the security groups do

terraform/projects/app-ecs-instances/main.tf
```diff
@@ -47,6 +41,10 @@ variable "ecs_instance_ssh_keyname" {
   default     = "ecs-monitoring-ssh-test"
 }
 
+variable "ecs_optimised_ami_version" {
```
idavidmcdonald: This move to a version number rather than an AMI version, does it mean that the README instructions on finding the latest AMI ID is no longer so useful? (I don't know the answer)

setup.sh
```diff
@@ -123,11 +136,19 @@ apply () {
         if [ $DEV_ENVIRONMENT = 'true' -a "$1" = 'infra-networking' ] ; then
                 import_shared_dev_route53
         fi
-        aws-vault exec ${PROFILE_NAME} -- $TERRAFORMPATH apply --var-file=$TERRAFORMTFVARS --auto-approve
+
+        # store output in a subprocess
+        exec 5>&1
+        OUTPUTS=$(aws-vault exec ${PROFILE_NAME} -- $TERRAFORMPATH apply --var-file=$TERRAFORMTFVARS --auto-approve|tee >(cat - >&5))
```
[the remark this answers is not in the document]
kentsanggds: not being used, could take it out, as was planning on using it to check for errors.

setup.sh
```diff
@@ -123,11 +136,19 @@ apply () {
...
+        # TODO: capture Errors and then stop any other executions of scripts
```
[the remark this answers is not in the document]
kentsanggds: Not sure, I'll take out the comment as probably covers it

terraform/projects/infra-jump-instance/main.tf
```diff
@@ -0,0 +1,109 @@
...
+  default     = "ecs-monitoring"
+}
+
+variable "stack_name" {
+  type        = "string"
+  description = "Unique name for this collection of resources"
+  default     = "ecs-monitoring"
+}
+
+variable "enable_jumpbox" {
```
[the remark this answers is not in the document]
kentsanggds: no can take out

terraform/projects/infra-jump-instance/main.tf
```diff
@@ -0,0 +1,109 @@
...
+  name    = "jump"
+  type    = "A"
+  ttl     = 60
+
+  records = ["${module.jumpbox.jumpbox_ip}"]
+}
+
+## Outputs
+
+output "jump_box_ip" {
```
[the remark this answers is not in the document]
kentsanggds: ok, will make consistent

terraform/modules/common/ami/main.tf
kentsanggds: Can probably take out

terraform/modules/common/instance/main.tf
kentsanggds: I think it's only used by the jumpbox project, so maybe could move the code into jumpbox?

setup.sh
```diff
@@ -152,6 +176,33 @@ taint() {
         aws-vault exec ${PROFILE_NAME} -- $TERRAFORMPATH taint $2
 }
 
+jumpbox() {
+        if [ $DEV_ENVIRONMENT != 'true' ] ; then
+                echo "Jumpbox is only available for dev environments"
+                return
+        fi
+
+        EC2_DATA=$(aws-vault exec $PROFILE_NAME -- aws ec2 describe-instances --filters "Name=tag:Environment,Values=$ENV" "Name=instance-state-name,Values=running")
```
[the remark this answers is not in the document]
kentsanggds: I'm not sure it's necessary to add additional comments, the code is fairly readable as it's split up into fairly atomic parts

terraform/modules/common/instance/main.tf
kentsanggds: can do, but there are a few places where this is not done in the codebase

terraform/modules/common/instance/main.tf
kentsanggds: Think this module should be part of jumpbox project

terraform/modules/common/ssh_key/main.tf
kentsanggds: This is also only used by the jumpbox so could move to that project

terraform/modules/common/ami/main.tf
kentsanggds: Think so, though  I'm not sure if there needs to be any change as it's specific to the amazon release

terraform/projects/infra-jump-instance/main.tf
kentsanggds: Don't think this is used

terraform/modules/common/ami/main.tf
kentsanggds: I'll have a chat with Ed to see what he thinks

terraform/modules/jumpbox/main.tf
kentsanggds: ok

terraform/projects/infra-jump-instance/README.md
kentsanggds: Don't think these instructions are used, there are instructions in the main `README` that explains how to use the jumpbox

terraform/projects/app-ecs-instances/main.tf
kentsanggds: README should probably be updated? Will check with Ed

terraform/modules/common/ssh_key/main.tf
kentsanggds: Could be in the jumpbox project

idavidmcdonald:
Oh and did we ever get to the bottom of why a new jumpbox is created everytime you run make apply even if you haven't made any code changes? Would be good to avoid that (although I'd be willing to let it slide and put it as a bug ticket given how long this PR has been open)

setup.sh
```diff
@@ -123,11 +136,17 @@ apply () {
         if [ $DEV_ENVIRONMENT = 'true' -a "$1" = 'infra-networking' ] ; then
                 import_shared_dev_route53
         fi
+
+        # store output in a subprocess
```
idavidmcdonald: I think this comment is no longer needed?

setup.sh
```diff
@@ -138,6 +157,7 @@ destroy () {
                 remove_shared_dev_route53
         fi
 
+        # store output in a subprocess
```
idavidmcdonald: I think this one is also not needed?

setup.sh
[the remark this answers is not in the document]
idavidmcdonald: Cool, happy to go with that then

terraform/projects/app-ecs-albs/main.tf
```diff
@@ -128,7 +128,10 @@ resource "aws_acm_certificate" "monitoring_cert" {
 
 resource "aws_route53_record" "monitoring_cert_validation" {
   # Count matches the domain_name plus each `subject_alternative_domain`
-  count = "${1 + length(concat(aws_route53_record.prom_alias.*.fqdn, aws_route53_record.alerts_alias.*.fqdn))}"
+  # This was originally the sum of the alerts.fqdn and prom.fqdns which was nice and self documenting
```
idavidmcdonald: Would be good to get your other PR merged and this rebased out of here as it conflicts

terraform/projects/infra-jump-instance/main.tf
```diff
@@ -0,0 +1,103 @@
...
+## Resources
+
+module "jumpbox" {
+  source          = "../../modules/jumpbox"
+  aws_region      = "${var.aws_region}"
+  subnet_id       = "${element(data.terraform_remote_state.infra_networking.public_subnets, 0)}"
+  stack_name      = "${var.stack_name}"
+  vpc_id          = "${data.terraform_remote_state.infra_networking.vpc_id}"
+  security_groups = ["${data.terraform_remote_state.infra_security_groups.monitoring_external_sg_id}"]
+  allowed_cidrs   = "${var.jumpbox_cidrs}"
```
idavidmcdonald: Are we allowing any IP address access to the jumpbox as opposed to just the GDS IPs here which would have been the default if we had not provided a value for `allowed_cidrs`? If so, is there a reason for that? I can't remember how cidr blocks work exactly so I might have misunderstood

terraform/projects/infra-jump-instance/README.md
```diff
@@ -0,0 +1,56 @@
+## Project: infra-jump-instance
```
idavidmcdonald: It appears these docs have been written manually meaning running `make docs` will overwrite them. Instead they can be included at the top in a docblock of the project file and they will be produced using `make docs`.

Makefile
```diff
@@ -49,14 +53,26 @@ apply-single: ## Apply terraform for a project, make apply-single project=<proje
 destroy-single: ## Destroy terraform for a project, make destroy-single project=<project name>
 	. ./setup.sh -d ${project}
 
-.PHONY: destroy
-destroy: ## Destroy stack
-	. ./setup.sh -d
+.PHONY: init-list
+init-list: ## Init terraform for a list of projects, make init-list list="<comma separated list of projects>"
```
idavidmcdonald: Any chance we could put the make list stuff in as a separate PR just for ease of keeping this PR small? I haven't had a chance to look at that new code/test it

setup.sh
kentsanggds: 👍

setup.sh
kentsanggds: 👍

terraform/projects/app-ecs-albs/main.tf
kentsanggds: 👍

terraform/projects/infra-jump-instance/main.tf
kentsanggds: I'll check with @ejrowley on this

Makefile
kentsanggds: Sure, will create a new PR for this

terraform/projects/infra-jump-instance/README.md
kentsanggds: ok

kentsanggds:
Had a chat with @ejrowley and it appears that by adding  `security_groups` to `ignore_changes`, it should prevent the instance from being destroyed.

PR #19, opened 2018-05-24
Change user permissions on instance user data template

DavidJeche:
This allows prometheus to write to the prometheus data directory but blocks public access.

https://trello.com/c/Pc4zwLxe/419-review-our-use-of-chmod-777-in-user-data

### DIFF
--- a/terraform/projects/app-ecs-instances/instance-user-data.tpl
+++ b/terraform/projects/app-ecs-instances/instance-user-data.tpl
@@ -32,10 +32,20 @@ else
   echo "[$(date '+%H:%M:%S %d-%m-%Y')] attach-volume: /dev/$DEVICE is already formatted: $(file -s /dev/$DEVICE)"
 fi
 
-# Allow prometheus container user access to read/write/execute within container
+
+#Mount volume to be used by prometheus container
 mkdir -p /ecs/prometheus_data
 mount /dev/$DEVICE /ecs/prometheus_data
-chmod 777 /ecs/prometheus_data
+
+
+
+#Create prometheus group and allow it to read and write to our volume for storing prometheus data. Note, 65534 is 
+#chosen as the UID to be added to the prometheus group as this is the UID that prometheus in the docker container runs as.
+
+groupadd --system --gid 65534 prometheus
+useradd --system --uid 65534 --gid 65534 prometheus
+chown prometheus:prometheus /ecs/prometheus_data
+chmod -R 760 /ecs/prometheus_data
 
 # Set any ECS agent configuration options
 echo 'ECS_CLUSTER=${cluster_name}' >> /etc/ecs/ecs.config


terraform/projects/app-ecs-instances/instance-user-data.tpl
```diff
@@ -32,10 +32,15 @@ else
   echo "[$(date '+%H:%M:%S %d-%m-%Y')] attach-volume: /dev/$DEVICE is already formatted: $(file -s /dev/$DEVICE)"
 fi
 
-# Allow prometheus container user access to read/write/execute within container
+# Allow prometheus container user access to read/write metrics to persistent volume mount from within a container
 mkdir -p /ecs/prometheus_data
+
+#Cretaing prometheus & granting access to prometheus data files. We create the group & match it to the gid
```
philandstuff: can this be simple present tense "create" and "grant" rather than "creating" and "granting"

terraform/projects/app-ecs-instances/instance-user-data.tpl
```diff
@@ -32,10 +32,15 @@ else
...
+groupadd --system --gid 65534 prometheus
+useradd --uid 65534 --gid 65534 --system --groups prometheus
 mount /dev/$DEVICE /ecs/prometheus_data
-chmod 777 /ecs/prometheus_data
+chown prometheus:prometheus /ecs/prometheus
+chmod -R 760 /ecs/prometheus_data
```
philandstuff: I don't understand this change. You've created a `prometheus` user and group, but how does that relate to the prometheus docker container? We never specify a user or group for it to run as.

philandstuff: Also, don't you need the execute bit on the directory, in order to `cd` into it? shouldn't this be `770` not `760`?

philandstuff: Where do the UID/GID come from? Can you link to some docs/Dockerfile source or something to show why this works?

idavidmcdonald: I was wondering about where the UID/GID come from too. I was of the understanding that it runs as `nobody` and therefore is `99/99` as per https://github.com/prometheus/prometheus/issues/3441

DavidJeche: @idavidmcdonald This is a value that is static within the docker container. We use this value in order to set a predictable UID/GID. 

We have the option to possibly set it manually within the container however this will add unnecessary complexity. I believe this is the best method thus far.

terraform/projects/app-ecs-instances/instance-user-data.tpl
```diff
@@ -32,10 +32,15 @@ else
   echo "[$(date '+%H:%M:%S %d-%m-%Y')] attach-volume: /dev/$DEVICE is already formatted: $(file -s /dev/$DEVICE)"
 fi
 
-# Allow prometheus container user access to read/write/execute within container
+# Allow prometheus container user access to read/write metrics to persistent volume mount from within a container
 mkdir -p /ecs/prometheus_data
+
+#Cretate prometheus & grant access to prometheus data files. We create the group & match it to the gid
+groupadd --system --gid 65534 prometheus
+useradd --uid 65534 --gid 65534 --system --groups prometheus
```
idavidmcdonald: Glancing at the docs (https://linux.die.net/man/8/useradd), you might be able to do the groupadd and useradd in one line using `useradd --usergroup`. Maybe worth a quick look (and a double check that is correct) but not fussed if you don't use it (you might also feel it impacts the readability)

idavidmcdonald: Is the `--groups` option doing nothing here? We are already specifying that it should join gid `65534` which is the prometheus group?

We could also match the order of the above statement so that `--system` comes first just for prettiness 💄

terraform/projects/app-ecs-instances/instance-user-data.tpl
```diff
@@ -32,10 +32,15 @@ else
   echo "[$(date '+%H:%M:%S %d-%m-%Y')] attach-volume: /dev/$DEVICE is already formatted: $(file -s /dev/$DEVICE)"
 fi
 
-# Allow prometheus container user access to read/write/execute within container
+# Allow prometheus container user access to read/write metrics to persistent volume mount from within a container
 mkdir -p /ecs/prometheus_data
+
+#Cretate prometheus & grant access to prometheus data files. We create the group & match it to the gid
```
idavidmcdonald: Would be good to be explicit that the prometheus container runs as nobody which has UID `65534` by default (with a link to prove that if easy to find)

terraform/projects/app-ecs-instances/instance-user-data.tpl
idavidmcdonald: I'm also a tiny bit confused by the useradd. Does this user not already exist (`nobody` running as `65534`)? Are we trying to recreate that user here when we may not need to (and we could just attach it to the prometheus group?). Maybe I'm just not understanding the docs though

terraform/projects/app-ecs-instances/instance-user-data.tpl
```diff
@@ -32,10 +32,15 @@ else
...
+groupadd --system --gid 65534 prometheus
+useradd --uid 65534 --gid 65534 --system --groups prometheus
 mount /dev/$DEVICE /ecs/prometheus_data
```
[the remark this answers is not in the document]
DavidJeche: Yes we can do this. I should have to no impact on the outcome, it is just a preference/style/readability thing

terraform/projects/app-ecs-instances/instance-user-data.tpl
DavidJeche: You can not set GID this way. This is why we had to do this in this way.

DavidJeche: Yep, to be fair I just added that in there for readability. We can take this out since the groupadd shows this.

terraform/projects/app-ecs-instances/instance-user-data.tpl
DavidJeche: I don't know what you mean by link in this context. If you mean that this is the correct method I would say: 

'docker run prom/prometheus'

in a new terminal: 

docker exec -it <<hash of container>> /bin/sh
/prometheus $ whoami
nobody
/prometheus $ id
uid=65534(nobody) gid=65534(nogroup)

Through this you can verify the permissions. If this is not what you mean by link please clarify.

terraform/projects/app-ecs-instances/instance-user-data.tpl
DavidJeche: The name of the user can be anything, the UID is a bit different. nobody user will exists on our system but this user will not have the same UID as the one that exists in the container. 

I would get an error if I tried to create a user on the node & that user already existed. In this step what I do is I create the prometheus & I give it a custom UID 65534. 

When the volume(filesystem folder) is mounted in to the container it pretty much keeps host filesystem defaults. This means that in the container in order to be able to perform operations on the volume you need to have a matching UID as the user that prometheus runs as.

terraform/projects/app-ecs-instances/instance-user-data.tpl
```diff
@@ -32,10 +32,16 @@ else
   echo "[$(date '+%H:%M:%S %d-%m-%Y')] attach-volume: /dev/$DEVICE is already formatted: $(file -s /dev/$DEVICE)"
 fi
 
-# Allow prometheus container user access to read/write/execute within container
+# Allow prometheus container user access to read/write metrics to persistent volume mount from within a container
```
idavidmcdonald: It might be a little clearer if the comment on this line was something like:
`Mount volume to be used by prometheus container`

And the comment on the second section was like:
` Create prometheus group and allow it to read and write to our volume for storing prometheus data. Note, 65534 is chosen as the UID to be added to the prometheus group as this is the UID that prometheus in the docker container runs as.`

My point regarding a link specifying the UID was that someone looking at this code probably won't know where the number 65534 has come from. They could find out by going into the docker container as you demonstrated but if they could just read it in some external docs for prometheus that would be quicker/easier. However I don't think the prometheus docs document this fact (the best I could find was github issues of people complaining about the choice of UID and referenced it there).

terraform/projects/app-ecs-instances/instance-user-data.tpl
```diff
@@ -32,10 +32,16 @@ else
...
 mkdir -p /ecs/prometheus_data
 mount /dev/$DEVICE /ecs/prometheus_data
-chmod 777 /ecs/prometheus_data
+
+
+#Create prometheus & grant access to prometheus data files. We create the group & match it to the gid
+groupadd --system --gid 65534 prometheus
+useradd --system --uid 65534 --gid 65534
+chown prometheus:prometheus /ecs/prometheus_data
```
idavidmcdonald: `prometheus:prometheus` refers to user with name `prometheus` and group with name `prometheus` if I understand this correctly. Do we have a user with name `prometheus`? We set up the group but I don't know if we have the user?

terraform/projects/app-ecs-instances/instance-user-data.tpl
DavidJeche: True, unfortunately we don't have any form of documentation around this. This is just one of those things. 

We will have to assume that the comments & the code will be enough for them. Let us also hope that they stumble upon these comments.

terraform/projects/app-ecs-instances/instance-user-data.tpl
DavidJeche: Ok I took the user out for some strange reason, however I have put the user back in and all should be well :)

terraform/projects/app-ecs-instances/instance-user-data.tpl
```diff
@@ -32,10 +32,22 @@ else
...
 mkdir -p /ecs/prometheus_data
 mount /dev/$DEVICE /ecs/prometheus_data
-chmod 777 /ecs/prometheus_data
+
+
+
+#Create prometheus group and allow it to read and write to our volume for storing prometheus data. Note, 65534 is 
+#chosen as the UID to be added to the prometheus group as this is the UID that prometheus in the docker container runs as.
+
+#Create prometheus & grant access to prometheus data files. We create the group & match it to the gid
```
idavidmcdonald: I don't think we need this line 46 anymore as 43 and 44 are fairly descriptive and feel like they are already saying this. Do you agree?

DavidJeche: Yep, I agree

PR #23, opened 2018-05-31
Make command to update docs

idavidmcdonald:
Note, doesn't add a check to our CI but that can be done as a later step


issyl0:
Yay, cross-team sharing!

PR #24, opened 2018-05-31
Set nice dev urls

idavidmcdonald:
https://trello.com/c/addX1GVK/413-nice-urls-for-prometheus (Bonus part of this ticket, staging/prod URLs already merged separately)

## What

Enabled the creation of nice urls for developers to be automatically created when creating a dev stack. 

- To enable nice urls we had to manually create the `dev.gds-reliability.engineering` DNS hosted zone and then import this as a resource in the `infra-networking-route53` project. 
- The dev stack doesn't get deployed on the staging and production environments and the staging and production hosted zones don't get deployed on the `gds-tech-ops` dev environment.
- In order to be able to add the new dev stack to `dev.gds-reliability.engineering` hosted zone we had to import it via `terraform import`, the shared dev DNS zone is hardcoded in the `setup.sh`.
- A decision was made to split out the `route53` resources into it's own project because it depended on outputs from the `app-ecs-albs` project.

Also fixed an issue around tearing down a dev stack as there was a `prevent_destroy` on the EBS volumes which we want for staging and production but not for a dev environment.

- in order to do this we had to manually introduce a policy `Prevent_delete_EC2_volume` for the Administrator roles in staging and production accounts as the Administrator account is not currently managed by terraform. We will need to revisit IAM roles and policies in the future to have this all managed within terraform.
- In order to ensure that we can destroy the `infra-networking-route53` project we had to use `terraform state rm` to remove the `shared_dev_subdomain` resource from the remote state bucket.

Refactored the setup.sh file to remove `USE_AWS_VAULT` flag as everyone in the RE Observation team is using `aws-vault` now.

In the `Makefile` and `setup.sh` more single actions were added to speed up the development cycle, and added a check to ensure that the stack backend is available before running terraform.

### DIFF
--- a/README.md
+++ b/README.md
@@ -46,12 +46,11 @@ Executing `make` on the command line will give you a list of possible commands t
 In order to create a new stack you can run these make commands in order:
 
 ```shell
-# ensure that you have set up and sourced your environment variables in `environment.sh`
+# ensure that you have set up and sourced your environment variables using `source environment.sh`
 
 make create-stack   # Create the terraform stack env vars
 make create-bucket  # Create the terraform state bucket
 make init           # Initialise terraform
-make plan           # Plan all terraform
 make apply          # Apply all terraform, auto approves
 ```
 
@@ -76,12 +75,11 @@ How to use the setup.sh shell script
 In order to create a new stack run the following commands in order:
 
 ```shell
-# ensure that you have set up and sourced your environment variables in `environment.sh`
+# ensure that you have set up and sourced your environment variables using `source environment.sh`
 
-. ./setup.sh -s     # create stack config files `backend` and `tfvars` 
+. ./setup.sh -s     # create stack config files `backend` and `tfvars`
 . ./setup.sh -b     # create the terraform bucket for holding the state
 . ./setup.sh -i     # initialise the terraform state
-. ./setup.sh -p     # plan terraform to see what will change in the stack
 . ./setup.sh -a     # apply terraform
 ```
 
--- a/README.md
+++ b/README.md
@@ -29,8 +29,7 @@ Before using the Makefile or shell script you will need to make a copy of the `e
 
 ```shell
 export TERRAFORM_BUCKET=<terraform state bucket name, should be unique or match `remote_state_bucket` in `tfvars` file for staging / production>
-export PROFILE_NAME=<if you are using `aws-vault` your profile name in `~/.aws/config` to access RE AWS>
-export USE_AWS_VAULT=<`true` if you are using `aws-vault`>
+export PROFILE_NAME=<your profile name in `~/.aws/config` to access RE AWS>
 export ENV=<your test environment, or `staging` / `production`>
 ```
 
--- a/README.md
+++ b/README.md
@@ -33,7 +33,7 @@ export PROFILE_NAME=<your profile name in `~/.aws/config` to access RE AWS>
 export ENV=<your test environment, or `staging` / `production`>
 ```
 
-Applying or destroying the stack on the staging and production environments has been blocked but is possible on other development environments.
+Applying or destroying the entire stack on the staging and production environments has been blocked but is possible on other development environments.
 
 <details>
 <summary>
@@ -95,6 +95,14 @@ To apply terraform for a particular project:
 `. ./setup.sh -a <project name in terraform/projects>`
 
 </details>
+
+Once you have deployed your development stack you should be able to reach the prometheus dashboard using this url pattern:
+
+`http://prom-1.<your test environment specified in the ENV environment variable>.dev.reliability.engineering`
+
+e.g.
+
+`http://prom-1.your-test-stack.dev.reliability.engineering`
 
 ## Development process
 
--- a/setup.sh
+++ b/setup.sh
@@ -5,7 +5,6 @@ TERRAFORMBACKVARS=$(pwd)/stacks/${ENV}.backend
 TERRAFORMTFVARS=$(pwd)/stacks/${ENV}.tfvars
 ROOTPROJ=$(pwd)
 TERRAFORMPROJ=$(pwd)/terraform/projects/
-USE_AWS_VAULT=${USE_AWS_VAULT}
 declare -a COMPONENTS=("infra-networking" "infra-security-groups" "app-ecs-instances" "app-ecs-albs" "infra-networking-route53"  "app-ecs-services")
 declare -a COMPONENTSDESTROY=("app-ecs-services" "infra-networking-route53" "app-ecs-albs" "app-ecs-instances" "infra-security-groups" "infra-networking")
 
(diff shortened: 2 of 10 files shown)


setup.sh
```diff
@@ -96,14 +138,36 @@ apply () {
         else
                 $TERRAFORMPATH apply --var-file=$TERRAFORMTFVARS --auto-approve
         fi
+
+        # For development stacks, for the `infra-networking-route53` project, ensure that
+        # the shared_dev_route53 resource has been imported into terraform before applying
```
idavidmcdonald: This comment I think is slightly wrong. Maybe you could help @kentsanggds. 

At the moment if you are applying the `infra-networking-route53` project for the first time you will
- apply
- import_shared_dev_route53
- apply again

why do we want to apply twice as opposed to just doing:
- import_shared_dev_route53
- apply again

setup.sh
```diff
@@ -96,14 +138,36 @@ apply () {
...
 }
 
 destroy () {
+# Destroy a terraform project
         echo $1
 
         cd $TERRAFORMPROJ$1
         if [ "${USE_AWS_VAULT}" = 'true' ] ; then
                 aws-vault exec ${PROFILE_NAME} -- $TERRAFORMPATH destroy --var-file=$TERRAFORMTFVARS --auto-approve
+                # What is this doing?
```
idavidmcdonald: Sorry, @kentsanggds could you just explain to me what this logic is doing but more so why it is doing it please?

setup.sh
kentsanggds: A state file needs to be present otherwise it won't do the import, hence create the state file, then import and apply

idavidmcdonald: Thanks for the clarity

setup.sh
```diff
@@ -28,21 +30,45 @@ fi
...
+
+create_stack_tfvars() {
+# Creates or updates .tfvars file for this stack in the stacks directory
+#
+# If function is not given an argument then will set tfvars `initialised` to "0".
+# This represents that the required terraform resources have not yet been imported.
+#
+# If function is given an argument then will update tfvars to set `initialised`
+# to that value. This will mean that the required terraform resources have been
+# imported and `terraform apply` can be run.
```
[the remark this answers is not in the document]
idavidmcdonald: Ah, I had a slightly different mental model in my head so this is helpful. I'll update the comment to more accurately describe the process.

setup.sh
kentsanggds: Do you mean `168`, when destroying we need to ignore `infra-networking-route53` errors otherwise it will always interrupt the destroy all as we can't ignore the `shared_dev` resource, see here for more detail - https://github.com/alphagov/prometheus-aws-configuration-beta/pull/24/commits/a6b7435107b37934aca2e02a803cf45207ae2b05

DavidJeche:
did anyone else see this warning when init project folders: 

Warning: output "dns_zone_id": must use splat syntax to access aws_route53_zone.subdomain attribute "zone_id", because it has "count" set; use aws_route53_zone.subdomain.*.zone_id to obtain a list of the attributes across all instances

DavidJeche:
At line 32 in setup.sh file I think we need to put ${ROOTPROJ} in the stacks cat creation. The file is not otherwise found <<EOF >stacks/${ENV}.tfvars

DavidJeche:
Ken, did you say you are going to leave out the set.sh for now since it does not effect deployment ?

setup.sh
```diff
@@ -42,70 +43,94 @@ fi
...
         cd $TERRAFORMPROJ$1
-        if [ "${USE_AWS_VAULT}" = 'true' ] ; then
-                aws-vault exec ${PROFILE_NAME} -- $TERRAFORMPATH destroy --var-file=$TERRAFORMTFVARS --auto-approve
-        else
-                $TERRAFORMPATH destroy --var-file=$TERRAFORMTFVARS
+
+        # For development stacks, for the `infra-networking-route53` project,
+        # remove the shared_dev_subdomain route53 state file
+        if [ $ENV != 'production' -a $ENV != 'staging' -a "$1" = 'infra-networking-route53' ] ; then
+                remove_shared_dev_route53
```
philandstuff: if I run `terraform destroy` manually, am I likely to delete the route 53 zone because I forget to unimport the shared zone?  Is there a way we can avoid this hazard?

setup.sh
```diff
@@ -42,70 +43,94 @@ fi
...
+# Creates versioned AWS bucket to store remote terraform state
+        aws-vault exec ${PROFILE_NAME} -- aws s3 mb "s3://${TERRAFORM_BUCKET}"
+        aws-vault exec ${PROFILE_NAME} -- aws s3api put-bucket-versioning  \
+        --bucket ${TERRAFORM_BUCKET} \
+        --versioning-configuration Status=Enabled
+}
+
+does_stack_config_exist() {
+        if [ -e "${ROOTPROJ}/stacks/${ENV}.backend" ] ; then
+                return "1"
```
philandstuff: it's confusing to use `1` to mean "true" in bash.. normally `0` is true and any nonzero value is false.

setup.sh
```diff
@@ -150,54 +173,75 @@ else
...
         -i) echo "Initialize terraform dir: ${ENV}"
-                if [ $2 ] ; then
-                        init $2
-                else
-                        for folder in ${COMPONENTS[@]}
-                        do
-                                init $folder
-                        done
+                does_stack_config_exist
+                if [ $? = 1 ] ; then
```
philandstuff: if you flip the `0` and `1` senses (see comment above in definition of `does_stack_config_exist`), then I think you can use it directly in the `if` block:

```sh
if does_stack_config_exist; then
```

setup.sh
```diff
@@ -150,54 +173,75 @@ else
...
+                                fi
                         fi
                 fi
         ;;
         *) echo "Invalid option"
         ;;
         esac
 fi
+
+cd $ROOTPROJ
```
philandstuff: I don't think this will do anything? [scripts are run inside a subshell, so changing directory in the subshell doesn't affect the caller](https://stackoverflow.com/a/255415)

terraform/projects/app-ecs-instances/main.tf
```diff
@@ -190,7 +190,13 @@ resource "aws_ebs_volume" "prometheus_ebs_volume" {
   type              = "gp2"
 
   lifecycle {
-    prevent_destroy = true
+    # This allows our EBS volumes to be destroyed, which is allowed behaviour
+    # for dev stacks but behaviour we want to avoid for staging or production.
+    # Therefore, we manually created a new policy in the staging and production
+    # account - `Prevent_delete_EC2_volume`, which has been attached to the
+    # AWS roles used by terraform to help prevent accidental EBS volume deletion
+    # https://github.com/hashicorp/terraform/issues/3116
```
philandstuff: this is an excellent comment

terraform/projects/infra-networking-route53/main.tf
```diff
@@ -57,12 +73,16 @@ data "terraform_remote_state" "app-ecs-albs" {
...
+# These resources are only created for staging or production environments (not dev)
 
 resource "aws_route53_zone" "subdomain" {
-  name = "${var.prometheus_subdomain}.gds-reliability.engineering"
+  count = "${local.create_staging_or_prod_count}"
+  name  = "${var.prometheus_subdomain}.gds-reliability.engineering"
 }
 
-resource "aws_route53_record" "prom-alias" {
+resource "aws_route53_record" "prom_alias" {
```
philandstuff: I think this record should be in the app-ecs-albs module; we shouldn't have a circular dependency between `infra-networking` and `app-ecs-albs` because, when creating a new environment, which one do you create first?

Specifically, I imagine that infra-networking creates the zone (with the zone id as an output), and app-ecs-elbs creates the record within the zone.

README.md
```diff
@@ -46,12 +45,11 @@ Executing `make` on the command line will give you a list of possible commands t
 In order to create a new stack you can run these make commands in order:
 
 ```shell
-# ensure that you have set up and sourced your environment variables in `environment.sh`
+# ensure that you have set up and sourced your environment variables using `source environment.sh`
 
 make create-stack   # Create the terraform stack env vars
 make create-bucket  # Create the terraform state bucket
 make init           # Initialise terraform
-make plan           # Plan all terraform
```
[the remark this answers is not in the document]
kentsanggds: We decided to remove it as the apply, will also run the plan and in the initial creating of a new stack there will be errors from the plan as resources are not available. So for a fresh stack we will need to apply each project so that the plans pass.

It is left in so that developers can see the impact of any changes without applying them in an existing stack.

setup.sh
kentsanggds: It doesn't destroy the shared `dev.gds-reliability.engineering` zone as terraform recognises that there are other resources in the zone. 

The un-import should happen automatically for dev stacks, and should allow the resource specific for a user to be removed without affecting other dev stacks.

setup.sh
kentsanggds: Thats true, will switch them around

setup.sh
kentsanggds: Tested and works a treat 👍

setup.sh
kentsanggds: ok will remove

terraform/projects/infra-networking-route53/main.tf
kentsanggds: ok, that could work, will then mean that we won't need a separate route53 terraform project

PR #25, opened 2018-05-31
Add alertmanager to ECS

DavidJeche:
Add Alert manager to prometheus & ensures communication is functional

Add alert manager to prometheus on ECS. Create a task, service & alb for prometheus. We then dynamically generate a prometheus config that has the alert manager name in it. Prometheus has two alerts that trigger at different times. These alerts are useful in order to test communication between the services.

There is DNS names created for all service. This is done via passing the URL down to app services. That is where the DNS name is associated to the LB & which is attached to the service.

[DNS adds are currently not fully automatic]


terraform/projects/app-ecs-services/alertmanager-service.tf
```diff
@@ -0,0 +1,123 @@
...
+    resources = ["arn:aws:s3:::${aws_s3_bucket.config_bucket.id}"]
+
+    actions = [
+      "s3:List*",
+    ]
+  }
+}
+
+#I need to refactor to ensure comppatability with other domains=
+resource "aws_route53_record" "alertmanager_dns" {
```
kentsanggds: Can this be moved to the infra-networking-route53 project please.

terraform/projects/app-ecs-services/prometheus-service.tf
```diff
@@ -66,6 +66,14 @@ data "aws_iam_policy_document" "prometheus_policy_doc" {
   }
 }
 
+resource "aws_route53_record" "prometheus_dns" {
```
kentsanggds: Can this be moved to the infra-networking-route53 project please.

terraform/projects/app-ecs-services/prometheus-service.tf
```diff
@@ -66,6 +66,14 @@ data "aws_iam_policy_document" "prometheus_policy_doc" {
...
+  zone_id = "${data.terraform_remote_state.infra-networking-route53.dns_zone_id}"
+  name    = "metrics.${var.stack_name}.gds-reliability.engineering"
```
kentsanggds: This should be `prom-1.monitoring.gds-reliability.engineering` for production, and `prom-1.monitoring-staging` for staging, `prom-1.<stack name>.dev.gds-reliability.engineering` for dev stacks

See - `prometheus_subdomain` in https://github.com/alphagov/prometheus-aws-configuration-beta/blob/master/stacks/production.tfvars

.gitignore
```diff
@@ -11,4 +11,6 @@ stacks/*
 environment.sh
 
 # editor config stuff
+.idea
```
JonathanHallam: was this part of your working this stack? If so should it be in your personal gitignore rather than here if there's no chance of it coming up again?

terraform/projects/app-ecs-services/alertmanager-service.tf
DavidJeche: I am not sure. The only problem I have with that is that I wanted the domain name to  be close to the service. I understand that we have sub-zone created within main.tf of route53 folder. If we keep the DNS record setting separate from service file. It means that when deleting services a stale DNS record will remain unless we perform a destroy on the DNS records folder. 

I think there should be some static values within the DNS folder, this will be something like prom-1 & alert-1. We should use the ability to set a custom value closer to the service. There is also an issue that if you run apply-single on the service dir it will not update infra-route53 folder.

I am happy to move this into that folder but I need you consider this point first.

terraform/projects/app-ecs-services/prometheus-service.tf
DavidJeche: See comment above

terraform/projects/app-ecs-services/prometheus-service.tf
DavidJeche: Ok this needs to be changed will look in to making this change.

terraform/projects/app-ecs-instances/main.tf
```diff
@@ -56,7 +56,7 @@ variable "ecs_instance_root_size" {
 variable "ecs_instance_ssh_keyname" {
   type        = "string"
   description = "SSH keyname for ECS container instances"
-  default     = "ecs-monitoring-ssh-test"
+  default     = "djeche-insecure"
```
JonathanHallam: probably shouldn't have your name in here

terraform/projects/app-ecs-instances/main.tf
```diff
@@ -146,6 +146,7 @@ data "template_file" "instance_user_data" {
 module "ecs_instance" {
   source = "terraform-aws-modules/autoscaling/aws"
 
+
```
JonathanHallam: extra whitespace

terraform/projects/app-ecs-instances/main.tf
```diff
@@ -184,13 +185,16 @@ module "ecs_instance" {
   )}"
 }
 
+
+
```
JonathanHallam: whitespace - `terraform fmt` should remove this?

terraform/projects/app-ecs-instances/main.tf
```diff
@@ -184,13 +185,16 @@ module "ecs_instance" {
...
 resource "aws_ebs_volume" "prometheus_ebs_volume" {
   availability_zone = "${element(data.terraform_remote_state.infra_networking.az_names, 0)}"
   size              = 500
   type              = "gp2"
 
   lifecycle {
-    prevent_destroy = true
+    prevent_destroy = false
```
JonathanHallam: changing this so you can rebuild in your dev environment is fine but this should be changed back for prod

terraform/projects/app-ecs-services/alertmanager-service.tf
DavidJeche: I ended up moving this, I think it is better for the time being until a use case comes up.

terraform/projects/app-ecs-services/prometheus-service.tf
DavidJeche: I removed this in favour of prometheus_subdomain

.gitignore
DavidJeche: I don't know my idea just keeps adding it. Ermmmmm that is consideration. I may need to set this up though.

terraform/projects/app-ecs-instances/main.tf
DavidJeche: changed, thanks for spotting that :D

terraform/projects/app-ecs-instances/main.tf
DavidJeche: removed

terraform/projects/app-ecs-instances/main.tf
DavidJeche: changed back to true :D

terraform/projects/app-ecs-albs/main.tf
```diff
@@ -129,19 +129,127 @@ resource "aws_lb_listener" "monitoring_external_listener" {
...
+    "${element(data.terraform_remote_state.infra_networking.public_subnets, 0)}",
+    "${element(data.terraform_remote_state.infra_networking.public_subnets, 1)}",
+    "${element(data.terraform_remote_state.infra_networking.public_subnets, 2)}",
+  ]
+
+  tags = "${merge(
+    local.default_tags,
+    var.additional_tags,
+    map("Stackname", "${var.stack_name}"),
+    map("Name", "${var.stack_name}-ecs-monitoring")
```
philandstuff: Is this the best Name for the ALB? Surely `"${var.stack_name}-alertmanager"` would be better?

terraform/projects/app-ecs-albs/main.tf
```diff
@@ -129,19 +129,127 @@ resource "aws_lb_listener" "monitoring_external_listener" {
...
+    protocol            = "HTTP"
+    healthy_threshold   = 2
+    unhealthy_threshold = 2
+    timeout             = "5"
+  }
+}
+
+
+resource "aws_lb_listener" "paas_proxy_listener" {
+  load_balancer_arn = "${aws_lb.alertmanager_external_alb.arn}"
```
philandstuff: why did you make this part of the alertmanager ALB? It's a bit weird as they don't have much to do with each other

terraform/projects/app-ecs-albs/main.tf
```diff
@@ -129,19 +129,127 @@ resource "aws_lb_listener" "monitoring_external_listener" {
...
 
-output "dns_name" {
+output "alertmanager_external_tg" {
+  value       = "${aws_lb_target_group.alertmanager_endpoint.arn}"
+  description = "External Alertmanager ALB target group"
+}
+
+output "prometheus_alb_zoneid" {
+  value = "${aws_lb.monitoring_external_alb.zone_id}"
+  description = "External Alertmanager ALB target group"
```
philandstuff: This description is inaccurate. Copypaste error?

terraform/projects/app-ecs-albs/main.tf
```diff
@@ -129,19 +129,127 @@ resource "aws_lb_listener" "monitoring_external_listener" {
...
+}
+
+output "alertmanager_alb_zoneid" {
+  value       = "${aws_lb.alertmanager_external_alb.zone_id}"
+  description = "External Alertmanager ALB target group"
```
philandstuff: Ditto

terraform/projects/app-ecs-albs/main.tf
```diff
@@ -129,19 +129,127 @@ resource "aws_lb_listener" "monitoring_external_listener" {
...
+}
+
+output "alertmanager_alb_dns" {
+  value       = "${aws_lb.alertmanager_external_alb.dns_name}"
+  description = "External Alertmanager ALB target group"
```
philandstuff: Ditto

terraform/projects/app-ecs-albs/main.tf
```diff
@@ -129,19 +129,127 @@ resource "aws_lb_listener" "monitoring_external_listener" {
...
+}
+
+output "prometheus_alb_dns" {
   value       = "${aws_lb.monitoring_external_alb.dns_name}"
-  description = "External Monitoring ALB dns_name"
+  description = "External Alertmanager ALB target group"
```
philandstuff: Ditto

terraform/projects/app-ecs-albs/main.tf
```diff
@@ -129,19 +129,127 @@ resource "aws_lb_listener" "monitoring_external_listener" {
...
 }
 
-output "zone_id" {
-  value       = "${aws_lb.monitoring_external_alb.zone_id}"
-  description = "External Monitoring ALB hosted zone ID"
+output "paas_proxy_alb_zoneid" {
+  value       = "${aws_lb.alertmanager_external_alb.zone_id}"
+  description = "External Alertmanager ALB target group"
```
philandstuff: Ditto

terraform/projects/app-ecs-albs/main.tf
```diff
@@ -129,19 +129,127 @@ resource "aws_lb_listener" "monitoring_external_listener" {
...
 }
+
+output "paas_proxy_alb_dns" {
+  value       = "${aws_lb.alertmanager_external_alb.dns_name}"
+  description = "External Alertmanager ALB target group"
```
philandstuff: Ditto

terraform/projects/app-ecs-services/config/vhosts/paas-proxy.conf
```diff
@@ -8,8 +8,16 @@ server {
...
+    return 200 "Static health check";
+  }
+
   resolver 8.8.8.8 valid=10s;
 
   # the AWS ip should work but is not resolving
   # resolver 10.0.0.2 valid=10s;  
 }
+
+
```
philandstuff: stray whitespace

terraform/projects/app-ecs-services/prometheus-service.tf
philandstuff: @kentsanggds I don't think `infra-networking-route53` should own all route53 records. That's not a good way of structuring our projects.

DavidJeche: I have moved this to the route53 folder. I will wait until we move it to networking, then I will add the records back into the services section.

terraform/projects/app-ecs-albs/main.tf
DavidJeche: map("Name", "${var.stack_name}-internal-alb")

I have renamed it to this, since we have multiple services on the alb that exposed internally. When we go multi-AZ we will need to add the the count to the alb name or AZ association.

terraform/projects/app-ecs-albs/main.tf
DavidJeche: This was inline with digram so basically they can share an ALB as long we as we correctly define the forwarding rules  They share one internal ALB in the AZ. They are currently only routed based on port number. PaaS proxy is on port 8080 & Alertmanager is on 80.

terraform/projects/app-ecs-albs/main.tf
DavidJeche: Yep... I hate typing when I can copy & reword :( 

I wanted to make it work, will clean these up to ensure the description matches the output 👍

setup.sh
```diff
@@ -56,17 +82,31 @@ create_bucket() {
...
         else
                 echo "$1 .terraform not found"
         fi
 }
 
+import_shared_dev_route53 () {
+# Imports into terraform the shared development route 53 zone that had not been set up
+# using Terraform
+        echo "import shared dev route53"
+        cd "$TERRAFORMPROJ/infra-networking-route53"
```
MuriloDalRi: This will need to change

setup.sh
```diff
@@ -96,14 +138,36 @@ apply () {
...
 }
 
 destroy () {
+# Destroy a terraform project
         echo $1
 
         cd $TERRAFORMPROJ$1
         if [ "${USE_AWS_VAULT}" = 'true' ] ; then
                 aws-vault exec ${PROFILE_NAME} -- $TERRAFORMPATH destroy --var-file=$TERRAFORMTFVARS --auto-approve
+                # What is this doing?
```
MuriloDalRi: What is this doing?

setup.sh
```diff
@@ -96,14 +138,36 @@ apply () {
         else
                 $TERRAFORMPATH apply --var-file=$TERRAFORMTFVARS --auto-approve
         fi
+
+        # For development stacks, for the `infra-networking-route53` project, ensure that
+        # the shared_dev_route53 resource has been imported into terraform before applying
+        if [ $ENV != 'production' -a $ENV != 'staging' -a "$1" = 'infra-networking-route53' ] ; then
```
MuriloDalRi: I think we might want to get rid of `infra-networking-route53` here

PR #26, opened 2018-06-01
BAU: Add swp and DS_Store files to git ignore

ejrowley:


philandstuff:
I'm yolo-merging this since it doesn't require a deploy

PR #27, opened 2018-06-01
Make the autoscaling group build 3 instances across 3 AZ

ejrowley:
This commit:
- increases the size of the autoscaling group to 3
- it passes all of the subnets to the autoscaling group
- creates 3 EBS volumes, 1 in each availability zone
- alters the cloud init script to query against 3 volume-ids to find the
volume created in the correct AZ

### DIFF
--- a/terraform/projects/app-ecs-services/templates/auth-proxy.conf.tpl
+++ b/terraform/projects/app-ecs-services/templates/auth-proxy.conf.tpl
@@ -18,7 +18,7 @@ server {
     return 200 "Static health check";
   }
 
-  resolver 8.8.8.8 valid=10s;
+  resolver 10.0.0.2 valid=10s;
 }
 
 server {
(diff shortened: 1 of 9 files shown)


DavidJeche:
This all looks good an  I am pretty much happy to approve. I am just wondering what happens if you change the min/max instance size to something more than number of subnets that we have :/

terraform/projects/app-ecs-services/templates/auth-proxy.conf.tpl
```diff
@@ -18,7 +18,7 @@ server {
     return 200 "Static health check";
   }
 
-  resolver 8.8.8.8 valid=10s;
+  resolver 10.0.0.2 valid=10s;
```
philandstuff: we can DRY this up by specifying `resolver` at the higher level - it's a valid `http` directive, so doesn't have to be inside a `server` block

terraform/projects/app-ecs-services/templates/auth-proxy.conf.tpl
```diff
@@ -27,15 +27,56 @@ server {
...
+}
+
+server {
+  listen 9090;
+  auth_basic "Prometheus";
+  auth_basic_user_file /etc/nginx/conf.d/.htpasswd;
```
philandstuff: ditto for the `auth_basic` directives

terraform/projects/app-ecs-services/templates/auth-proxy.conf.tpl
```diff
@@ -27,15 +27,56 @@ server {
   auth_basic_user_file /etc/nginx/conf.d/.htpasswd;
 
 
-  server_name ~(alert*);
+  server_name alerts-1.*;
 
+  set $alert "${alertmanager_1_dns_name}";
```
philandstuff: why do we use `set` here rather than just proxying directly to `${alertmanager_1_dns_name}`?

PR #31, opened 2018-06-08
Alertmanager pagerduty

idavidmcdonald:
https://trello.com/c/dljFeMNm/426-integrate-alertmanger-with-pagerduty-to-enable-alerts-to-become-pages

PR relies on https://github.com/alphagov/reliability-engineering-credentials/pull/11

- This adds integration with a re-tools team wide pager duty account. Alertmanager will send an alert to pagerduty when an alert is triggered on our staging or production Prometheis. Alerts will not be triggered on development stacks. Staging and production Prometheis are sharing a pagerduty service, this was done for speed but can be changed later if we need a lager degree of separation.

- We have added a `pass` terraform integration which means deploys will now rely on getting secrets out of the `reng-pass` in order to deploy a stack. This seems to work OK but we may need to investigate some potential issues around the pass store not being unlocked by a passphrase mid execution of terraform and needing to be done manually beforehand.

- Now we are dealing with "secrets" in our config files, we have tightened required encryption on the `config_bucket`.

- We've also removed hardcoding of checks that the environment is `staging` or `production` from the codebase and instead rely on the `DEV_ENVIRONMENT` environment variable. Note, after this PR is merged, developers will need to update any `tfvars` files they are using to include this new variable.


terraform/projects/app-ecs-services/alertmanager-service.tf
```diff
@@ -106,9 +106,25 @@ resource "aws_ecs_service" "alertmanager_server" {
...
+
 resource "aws_s3_bucket_object" "alertmanager" {
-  bucket = "${aws_s3_bucket.config_bucket.id}"
-  key    = "alertmanager/alertmanager.yml"
-  source = "config/alertmanager.yml"
-  etag   = "${md5(file("config/alertmanager.yml"))}"
+  bucket                 = "${aws_s3_bucket.config_bucket.id}"
+  key                    = "alertmanager/alertmanager.yml"
+  content                = "${var.dev_environment == "true" ? data.template_file.alertmanager_dev_config_file.rendered : data.template_file.alertmanager_config_file.rendered}"
+  server_side_encryption = "AES256"
```
philandstuff: why has the `etag` gone? can we still ensure that terraform won't re-upload the config file each time?

terraform/projects/app-ecs-services/main.tf
```diff
@@ -51,6 +57,11 @@ provider "template" {
   version = "~> 1.0.0"
 }
 
+provider "pass" {
+  store_dir     = "~/.reng-pass"
+  refresh_store = false          # do not call `git pull`
```
philandstuff: can this comment explain why we don't want to call `git pull`?

terraform/projects/app-ecs-services/main.tf
```diff
@@ -109,5 +110,44 @@ resource "aws_s3_bucket" "config_bucket" {
...
+      "Resource": "arn:aws:s3:::${aws_s3_bucket.config_bucket.id}/*",
+      "Condition": {
+        "Null": {
+          "s3:x-amz-server-side-encryption": "true"
+        }
+      }
+    }
+  ]
+}
+POLICY
```
philandstuff: 👍

terraform/projects/app-ecs-services/templates/alertmanager.tpl
```diff
@@ -0,0 +1,10 @@
+global:
+  resolve_timeout: 5m
+
+route:
+  receiver: 'pagerduty'
+
+receivers:
+- name: 'pagerduty'
```
philandstuff: it upsets me that terraform templates don't support conditionals 😢

terraform/projects/infra-networking/main.tf
```diff
@@ -39,9 +45,9 @@ locals {
     Project   = "infra-networking"
   }
 
-  create_dev_count          = "${(var.stack_name == "production" || var.stack_name == "staging") ? 0 : 1}"
+  create_dev_count          = "${var.dev_environment == "true" ? 1 : 0}"
```
philandstuff: is it worth inlining this variable now?

terraform/projects/app-ecs-services/alertmanager-service.tf
idavidmcdonald: We have done some investigation 🕵️ 

An Etag is generated automatically by AWS when calling `PutObject` (https://docs.aws.amazon.com/AmazonS3/latest/API/RESTCommonResponseHeaders.html) with encryption so we still get an Etag for our object. We've also checked the Etag of the encrypted object is identical to the Etag of the object uploaded without encryption.

If we add the etag back in as before then we get the following error message:
`Error: aws_s3_bucket_object.nginx-htpasswd: "etag": conflicts with server_side_encryption ("AES256")`.

`Etag` is not a compatible argument if we are defining `kms_key_id` as per the docs (https://www.terraform.io/docs/providers/aws/r/s3_bucket_object.html#etag) but we aren't using `kms` to encrypt and instead using AES256 as our encryption method so we are surprised to see this error message. Maybe is worth us opening an issue with Terraform unless you can think of some areas to further investigate.

We've manually testing running make apply twice and it appears that the files are not reuploaded again given that the "last modified" datetime does not update when you apply the second time.

We also aren't quite sure whether AWS or Terraform is responsible for not reuploading a file if it matches the Etag and haven't been able to find that out either.

idavidmcdonald: We've checked and if the file stays the same then terraform does not try and reupload it on top of the existing file.

terraform/projects/app-ecs-services/main.tf
```diff
@@ -109,5 +110,44 @@ resource "aws_s3_bucket" "config_bucket" {
   }
 }
 
+resource "aws_s3_bucket_policy" "config_bucket_policy" {
```
idavidmcdonald: Terraform thinks this object is changing everytime we run make apply which it shouldn't so we should fix this.

README.md
```diff
@@ -23,14 +23,38 @@ You should end up with something similar to this in your `.aws/config` file:
...
 
+### Set up the `terraform-provider-pass` third-party Terraform plugin
+
+This acts as middleware between terraform and the reng-pass password store and enables us to pass secrets into terraform.
+
+install dep
+  `brew install dep`
+
+Download the source code
+  `go get github.com/camptocamp/terraform-provider-pass`
```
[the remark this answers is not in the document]
philandstuff: I got something working:

```
go get github.com/camptocamp/terraform-provider-pass
GOBIN=~/.terraform.d/plugins/darwin_amd64 go install github.com/camptocamp/terraform-provider-pass
```

and the provider can now be used 😄

terraform/projects/app-ecs-services/main.tf
idavidmcdonald: We fixed this by adding the bucket policy version.

README.md
```diff
@@ -23,14 +23,24 @@ You should end up with something similar to this in your `.aws/config` file:
     role_arn=arn:aws:iam::<account-number>:role/<iam-role-name>
     mfa_serial=arn:aws:iam::<iam-user-id>:mfa/<iam-user-name>
 
+### Set up the `terraform-provider-pass` third-party Terraform plugin
+
+This acts as middleware between terraform and the reng-pass password store and enables us to pass secrets into terraform.
+
+  `go get github.com/camptocamp/terraform-provider-pass`
```
philandstuff: can this use a ``` block

setup.sh
```diff
@@ -116,7 +117,7 @@ apply () {
 
         # For development stacks, for the `infra-networking` project, ensure that
         # the shared_dev_route53 resource has been imported into terraform before applying
-        if [ $ENV != 'production' -a $ENV != 'staging' -a "$1" = 'infra-networking' ] ; then
+        if [ $DEV_ENVIRONMENT = 'true' -a "$1" = 'infra-networking' ] ; then
```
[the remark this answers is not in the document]
philandstuff: wait, we never actually added this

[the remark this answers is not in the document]
idavidmcdonald: The other blocker which was more going to take up some time learning how to fix it was that we refer to `$2` in our code a lot in reference to a second parameter passed into the script, however we don't usually pass a second parameter and it's optional. However if `$2` is not defined then our script fails so we need to add this ability to have `set -eu`. It took more than about 5 minutes of investigation with no immediate answer and we thought it wasn't essential to this PR so we stopped. Happy to look into it a bit more though.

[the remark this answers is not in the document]
philandstuff: 👍 this is a good explanation

PR #32, opened 2018-06-08
Recommend people use tfenv

philandstuff:
The current README recommends running

  brew install terraform@0.11.7

which doesn't actually work because there's no formula named
terraform@0.11.7.  The way homebrew works is that @-symbols are just
part of the name of a formula: if multiple versions are provided, it's
because someone is maintaining a separate formula manually.  This is
useful for things like mysql (which has mysql@5.5 and mysql@5.6) but
isn't implemented for terraform.  You can see these @-suffixed
formulas listed when you run `brew search <formula>` (eg `brew search
mysql` or `brew search terraform`).

It's worth pinning our terraform version though.  For this, we can use
`tfenv` (used by govuk infra and recommended by @issyl0), and using a
.terraform-version file to pin the particular terraform version.


DavidJeche:
This seems like an awesome idea. I have no issues with this. I looked at docs and it mentioned it will always use the latest unless otherwise specified.

PR #38, opened 2018-06-15
WIP: Deploy container instances in separate ASGs

philandstuff:
This commit splits the one-ASG-with-three-instances into
three-ASGs-with-one-instance-each.

The goal I am going for with this commit is: we need to be able to
tell each separate prometheus service
aws_ecs_service.prometheus_server to run on a separate instance.  I
would also like to make our code flexible enough that dev environments
can run one single instance, and staging and production could scale
down to 2, or up to 4 (with two prometheis in a single AZ), if desired.

I had considered telling the services to run with [placement contraints][1]
pinning each prometheus service to a specific availability zone.
However, if there are only 1 or 2 instances, then some of the AZs will
be empty, and pinning a service to that AZ will mean it can never run
at all.  Moreover, there's no clean way to know which AZs are
available.  Finally: if an instance in an ASG dies, will the instance
that replaces it be in the same AZ, or could it come back in a
different one?

I thought a better way of doing this would be to set an explicit
placement attribute in the ECS agent for each instance.  In order to
do this, each instance needs to be created by a separate ASG (because
an ASG has a single user data script, and the user data script sets up
the ECS agent settings).  Now we can pin prometheus services to
specific instances using the custom `prometheus-instance` attribute,
and have confidence that those specific instances exist.

Unfortunately, [terraform modules don't currently support the count attribute][2],
so the only way to have multiple ASGs using this module
is to copy/paste the module multiple times.  This feels like the
least-worst option to me.  (In future, to scale up we would need to
copy/paste more.  To make the number of instances configurable, we can
make each ASG conditional based on the desired number of instances.
For details, see the [module docs][3].)

This has a side benefit that we can now specify the exact volume id in
user data again, rather than doing the trick where we grab a list of
volumes from the API and select the one that's in the same AZ as
us.  (This also enables us to deploy multiple instances in the same AZ
in future if we want.)

In order to be certain that we're pairing up availability zones,
private subnets, ASGs and volume ids correctly, I introduced a
data.aws_subnet.private_subnets data source, and pull availability
zone config from there to configure the aws_ebs_volumes.  This means
we can be confident that the EBS volume is in the same AZ as the ASG
it's paired with.

Note that the number 3 is still hardcoded throughout
app-ecs-instances/main.tf; actually making things configurable will
come in future commits.

I'm issuing this as a PR on its own, even though it isn't the whole multi-AZ story, for a number of reasons:

 - it's big enough by itself and worth making visible
 - I want to deploy this change and ensure that the ECS attributes are present on the instances before deploying any changes that rely on those attributes

[1]: https://docs.aws.amazon.com/AmazonECS/latest/developerguide/task-placement-constraints.html
[2]: https://github.com/hashicorp/terraform/issues/953
[3]: https://registry.terraform.io/modules/terraform-aws-modules/autoscaling/aws/2.6.0

### DIFF
--- a/terraform/projects/app-ecs-albs/README.md
+++ b/terraform/projects/app-ecs-albs/README.md
@@ -19,11 +19,13 @@ Create ALBs for the ECS cluster
 |------|-------------|
 | alertmanager_alb_dns | External Alertmanager ALB DNS name |
 | alertmanager_alb_zoneid | External Alertmanager ALB zone id |
-| alertmanager_external_tg | External Alertmanager ALB target group |
+| alerts_private_record_fqdn | Alert Managers private DNS fqdn |
 | monitoring_external_tg | External Monitoring ALB target group |
+| monitoring_internal_tg | External Alertmanager ALB target group |
 | paas_proxy_alb_dns | Internal PaaS ALB DNS name |
 | paas_proxy_alb_zoneid | Internal PaaS ALB target group |
-| pass_proxy_tg | Paas proxy target group |
+| paas_proxy_private_record_fqdn | PaaS Proxy private DNS fqdn |
+| paas_proxy_tg | Paas proxy target group |
 | prometheus_alb_dns | External Monitoring ALB DNS name |
 | zone_id | External Monitoring ALB hosted zone ID |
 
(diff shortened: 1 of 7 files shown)


philandstuff:
To maybe make things clearer, this commit is laying the groundwork so that something like 4325c27 can pin ECS tasks to specific EC2 instances.

philandstuff:
Marking as WIP while I explore an alternative solution that doesn't involve splitting the ASG

philandstuff:
closing in favour of #40

PR #39, opened 2018-06-15
Add the targets_s3_bucket environment variable

JonathanHallam:
This allows us to have a different s3 bucket for staging, production and development

Part of - https://trello.com/c/Ptl9OvfJ/397-paas-service-discovery-should-give-separate-instances-unique-instance-labels-and-add-org-and-space-labels-to-metrics


stacks/production.tfvars
```diff
@@ -2,6 +2,7 @@ remote_state_bucket = "prometheus-production"
 stack_name = "production"
 dev_environment = "false"
 prometheus_subdomain = "monitoring"
+targets_s3_bucket = "gds-prometheus-targets"
```
deanwilson: Why does this one not have the `prod` suffix? ;)

JonathanHallam: You're meant to be on holiday! Also because it's the default behaviour, prod's so important it doesn't need a suffix

JonathanHallam: although obviously if this is confusing can change :)

setup.sh
```diff
@@ -33,8 +33,9 @@ else
 cat <<EOF >stacks/${ENV}.tfvars
 remote_state_bucket = "${TERRAFORM_BUCKET}"
 stack_name = "${ENV}"
-dev_environment = "${DEV_ENVIRONMENT}"
+dev_environment = "true"
 prometheus_subdomain = "${ENV}"
+targets_s3_bucket="gds-prometheus-targets-dev"
```
idavidmcdonald: Could we maybe store this string as a variable at the top of the script, something like `DEFAULT_DEV_TARGETS_S3_BUCKET`

terraform/projects/app-ecs-services/main.tf
```diff
@@ -14,7 +14,7 @@ variable "aws_region" {
 variable "dev_environment" {
   type        = "string"
   description = "Boolean flag for development environments"
-  default     = "true"
+  default     = "false"
```
idavidmcdonald: I can't think of any particular issues with this but I'd be interested to know if you had a particular reason for making this change?

JonathanHallam: Basically because all of the other variables have prod as standard, figured they should probably be the same

setup.sh
JonathanHallam: is this just to make it easier to change later on?

idavidmcdonald: My thinking was more that it's essentially a configuration value and it's good to have our config and code that does things live separately if possible (whether that be in different files or just separated at the top of the script). It makes it easier to see what all your config are without having to look in random places in the script.

PR #45, opened 2018-06-20
rejig how we manage paas-proxy and alertmanager private names

philandstuff:
This commit does a couple of things:

 - in the app-ecs-albs outputs, make it clear that
   alerts_private_record_fqdns is a list of FQDNs by giving it a
   plural name, and make it clear that paas_proxy_private_record_fqdn
   is a single FQDN by making a string instead of a list of exactly
   one string

 - in the prometheus configuration templating, restrict the number of
   alertmanagers we scrape and that we send alerts to, to the number
   of available_azs, so we don't try to scrape non-existant
   alertmanagers.


DavidJeche:
It looks good for side & everything is as described

PR #47, opened 2018-06-22
Update the way that the cert_validation gets it's count

kentsanggds:
When applying the terraform, errors were thrown due to terraform being unable to resolve the count. 
This changes the way the count is calculated.

PR #49, opened 2018-06-26
Ensure that an EC2 instance will die if not attached to an EBS volume

JonathanHallam:
https://trello.com/c/XaZpAVee/444-prevent-user-data-from-starting-if-an-ebs-volume-is-not-attached

This stops an instance from starting if it cannot attach to an EBS volume. This will mean that if a build fails it will do it in a way that we will notice and that we will not lose metrics. This fixes an issue whereby an instance would fail to attach to a volume because that volume had an old instance still attached.

### DIFF
--- a/terraform/projects/app-ecs-instances/instance-user-data.tpl
+++ b/terraform/projects/app-ecs-instances/instance-user-data.tpl
@@ -1,21 +1,40 @@
 #!/bin/bash
 # Attach EBS volume to instance
 echo "[$(date '+%H:%M:%S %d-%m-%Y')] installing dependencies for volume attaching"
-sudo yum install -y aws-cli wget
 
 REGION="${region}"
 DEVICE="xvdf"
-VOLUME_IDS="${volume_ids}"
 
 echo "[$(date '+%H:%M:%S %d-%m-%Y')] finding current instance ID"
-INSTANCE_ID="`wget -q -O - http://169.254.169.254/latest/meta-data/instance-id`"
+INSTANCE_ID=$(wget -q -O - http://169.254.169.254/latest/meta-data/instance-id)
 
 echo "[$(date '+%H:%M:%S %d-%m-%Y')] finding volume to attach"
 AZ="$(wget -q -O - http://169.254.169.254/latest/meta-data/placement/availability-zone)"
-VOLUME_ID="$(aws ec2 describe-volumes --filters Name=availability-zone,Values=$AZ --volume-ids $VOLUME_IDS --region $REGION --query Volumes[*].VolumeId --output text)"
+VOLUME_ID="$(aws ec2 describe-volumes --filters Name=availability-zone,Values="$AZ" --volume-ids ${volume_ids} --region "$REGION" --query Volumes[*].VolumeId --output text)"
 
-echo "[$(date '+%H:%M:%S %d-%m-%Y')] attaching volume: $VOLUME_ID"
-aws ec2 attach-volume --volume-id $VOLUME_ID --instance-id $INSTANCE_ID --device /dev/$DEVICE --region $REGION
+count=0
+DISK_AVAILABILITY="NA"
+until [ "$DISK_AVAILABILITY" = available ]; do
+    if [[ $count -le 10 ]]
+    then
+        sleep 10;
+        echo "Sleeping: waiting for volume to become available"
+        count=$((count+1));
+        DISK_AVAILABILITY=$(aws ec2 describe-volumes --region "$REGION" --filters Name=volume-id,Values="$VOLUME_ID" --query Volumes[0].State --output text)
+    else
+        break
+    fi
+done
+
+case $DISK_AVAILABILITY in
+        available)
+            echo "[$(date '+%H:%M:%S %d-%m-%Y')] attaching volume: $VOLUME_ID"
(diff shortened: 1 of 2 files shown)


terraform/projects/app-ecs-albs/main.tf
```diff
@@ -128,7 +128,7 @@ resource "aws_acm_certificate" "monitoring_cert" {
 
 resource "aws_route53_record" "monitoring_cert_validation" {
   # Count matches the domain_name plus each `subject_alternative_domain`
-  count = "${1 + length(concat(aws_route53_record.prom_alias.*.fqdn, aws_route53_record.alerts_alias.*.fqdn))}"
+  count = "${1 + (2 * length(data.terraform_remote_state.infra_networking.public_subnets))}"
```
ejrowley: This change should not be in this commit it's unrelated to the commit and there is another pull request which fixes the problem.

terraform/projects/app-ecs-instances/instance-user-data.tpl
```diff
@@ -1,21 +1,41 @@
 #!/bin/bash
 # Attach EBS volume to instance
 echo "[$(date '+%H:%M:%S %d-%m-%Y')] installing dependencies for volume attaching"
-sudo yum install -y aws-cli wget
+sudo yum install -y aws-cli wget jq
```
ejrowley: can we not add `jq` it is another dependancy to manage, instead use the --output text and POSIX tools (awk/sed). Reason being `jq` has had CVEs in the past which have taken along time to get fixed, I would not be surprised if the version being installed here had open CVEs against it

terraform/projects/app-ecs-albs/main.tf
DavidJeche: We understand but without this change the PR will not function as expected. The alternative PR didn't merge so we just took some of the logic.

terraform/projects/app-ecs-instances/instance-user-data.tpl
DavidJeche: We could not use jq and utilize the AWS cli, I got not issue with that. Everything has CVEs attached to it. Just make sure you stop people from getting to them :D 

We need to have a discussion about vuln management at some point.

DavidJeche: I have made the change and removed this from the script

terraform/projects/app-ecs-albs/main.tf
DavidJeche: I am not sure what you would like to do about this but it is up to you to fair.

terraform/projects/app-ecs-instances/instance-user-data.tpl
MuriloDalRi: It's still being installed in line 4 though

terraform/projects/app-ecs-instances/instance-user-data.tpl
```diff
@@ -1,23 +1,45 @@
...
+            echo "[$(date '+%H:%M:%S %d-%m-%Y')] attaching volume: $VOLUME_ID"
+            aws ec2 attach-volume --volume-id "$VOLUME_ID" --instance-id "$INSTANCE_ID" --device /dev/"$DEVICE" --region "$REGION";
+         ;;
+        *)
+            shutdown -h now;
+         ;;
+esac
 
 # Waiting for volume to finish attaching
+#I do not know why this does not work on the instance
```
MuriloDalRi: Is this working now?

terraform/projects/app-ecs-instances/instance-user-data.tpl
JonathanHallam: good spot, thanks, have removed

terraform/projects/app-ecs-instances/instance-user-data.tpl
JonathanHallam: yeah I've taken this line out about three times, kept coming up after rebasing, is gone for good now (hopefully)

terraform/projects/app-ecs-albs/main.tf
kentsanggds: I've updated the PR which should resolve this issue - https://github.com/alphagov/prometheus-aws-configuration-beta/pull/47, so that just needs to be approved and merged in, then you can rebase off master.

DavidJeche:
@ejrowley can you review this please :( & check to make sure you are happy with it

terraform/projects/app-ecs-instances/instance-user-data.tpl
```diff
@@ -1,21 +1,40 @@
...
-aws ec2 attach-volume --volume-id $VOLUME_ID --instance-id $INSTANCE_ID --device /dev/$DEVICE --region $REGION
+count=0
+DISK_AVAILABILITY="NA"
+until [ "$DISK_AVAILABILITY" = available ]; do
+    if [[ $count -le 10 ]]
+    then
+        sleep 10;
+        echo "Sleeping: waiting for volume to become available"
+        count=$((count+1));
+        DISK_AVAILABILITY=$(aws ec2 describe-volumes --region "$REGION" --filters Name=volume-id,Values="$VOLUME_ID" | jq -r '.Volumes[0].State')
```
ejrowley: We are still using jq here

PR #51, opened 2018-06-29
Make alert manager alert when there is 2 or less instances

DavidJeche:
This is an alert that will trigger when there is two or
less instances available on cluster.

I have tested and ensured functionality, I believe this will be routed to us based on the label tools.


terraform/projects/app-ecs-services/config/alerts/alerts.yml
```diff
@@ -10,3 +10,17 @@ groups:
     annotations:
         summary: "Service is taking longer than 2 seconds to respond"
         description: "The service name is {{ $labels.job }}. The URL under test is {{ $labels.instance }}"
+
+- name: AlertManager
+  rules:
+  - alert: AlertManager_Below_Threshold
+    expr: sum(up{job="alertmanager"}) by (job, instance) <= 2 and up{job="alertmanager"} == 0
```
ejrowley: Think this should be `<=1`

PR #55, opened 2018-07-04
Add alert on number of file_sd targets

philandstuff:
This adds an alert to ensure we are detecting some file_sd targets.

The query is a little odd (see comment).  To test it out, I checked
the query output in the prometheus UI until I got something that
fired during our recent outage.

This also changes the product name from "tools" to "prometheus" -- our
team name has changed, and will change again in future, but prometheus
will remain prometheus.

### DIFF
--- a/terraform/projects/app-ecs-services/config/alerts/alerts.yml
+++ b/terraform/projects/app-ecs-services/config/alerts/alerts.yml
@@ -5,7 +5,6 @@ groups:
     expr: up{job="alertmanager"} == 0 and on(job) sum by(job) (up{job="alertmanager"}) <= 1
     for: 10s
     labels:
-        severity: "P2"
         product: "prometheus"
     annotations:
         summary: "Service is below the expected instance Threshold"
@@ -19,7 +18,6 @@ groups:
     expr: (count(prometheus_sd_file_mtime_seconds) or count(up)*0) < 1
     for: 10s
     labels:
-        severity: "P2"
         product: "prometheus"
     annotations:
         summary: "No file_sd targets detected"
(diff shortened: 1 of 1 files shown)


terraform/projects/app-ecs-services/config/alerts/alerts.yml
```diff
@@ -1,14 +1,28 @@
...
+  - alert: NoFileSdTargets
+    # this expression is a little weird - count() has no value instead of 0 if there are no
+    # matching metrics.  But no value isn't less than 1 so we aren't able to trigger an alert
+    # that a regular count() is zero.  So we force missing values to be equal to
+    # zero using the `or count(up)*0` expression.
+    # This idea taken from: https://www.robustperception.io/existential-issues-with-metrics/
+    expr: (count(prometheus_sd_file_mtime_seconds) or count(up)*0) < 1
+    for: 10s
+    labels:
+        severity: "P2"
```
JonathanHallam: Do we have to add a severity here? Happy to if we do just a slight worry about doing enacting this without a shared understanding of severity levels

JonathanHallam: although it's just occurred to me that we're using P2 in other places too

philandstuff: yeah i think it was just here as an example of what labels might look like.  But we have product too, so i think we can kill severity until we define it better 🚮

philandstuff:
@JonathanHallam pushed an update, can you re-review?

JonathanHallam:
👍

PR #56, opened 2018-07-04
Specify --web.external-url when running Prometheus

philandstuff:
This commit updates our container definitions so that we can specify
the external URL that Prometheus can be accessed by.

This is important, because when Prometheus ships alerts to
Alertmanager, it tags them with a source URL, which by default is a
private non-routable URL.  To make our alerts more useful, we should
have a publicly-routable URL here instead.

Because each prometheus instance has a different public URL, they each
need a separate command-line parameter.  To achieve this I split the
single task definition into one per prometheus.


philandstuff:
Writing a comment to mollify the seal tomorrow

PR #58, opened 2018-07-10
Add an ASG schedule for dev stacks

kentsanggds:
## What

The schedule is based on UTC times, so during BST the schedule will start at 19:00 rather than 18:00, but this is probably ok for the moment.

The default list of cron schedules is based around the core hours of 9 - 18 hours, Monday to Friday.

As developers can spin up the instances outside of these core hours to ensure that the EC2 instances are terminated the scheduled EC2 termination will be scheduled for at each hour outside of the core hours.

If you have a different working pattern, you can add the `asg_dev_scaledown_schedules` variable in your tfvars files with a list of cron schedules.

## Reference

https://trello.com/c/Gn0fdyOr/447-destroy-our-dev-stacks-overnight-to-reduce-costs

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,7 +4,7 @@ Terraform configuration to manage a Prometheus server running on AWS.
 
 ## Deployment ##
 
-We track deployments in this spreadsheeet: https://docs.google.com/spreadsheets/d/1Kd390OLID_ZL8U9_wdoEJCkiDU3nSHsbKzovcRIUsW4/edit#gid=413288099
+We track deployments in this spreadsheet: https://docs.google.com/spreadsheets/d/1Kd390OLID_ZL8U9_wdoEJCkiDU3nSHsbKzovcRIUsW4/edit#gid=413288099
 Please follow the deployment checklist and keep the deployment log up to date
 
 ## Setup ##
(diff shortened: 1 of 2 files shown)


terraform/projects/app-ecs-instances/main.tf
```diff
@@ -185,6 +199,16 @@ module "ecs_instance" {
   )}"
 }
 
+resource "aws_autoscaling_schedule" "asg_dev_scaledown_schedules" {
+  scheduled_action_name  = "asg_dev_scaledown_schedule-${element(var.asg_dev_scaledown_schedules, count.index)}"
```
idavidmcdonald: Just checking that you wanted to have the name contain the cron expression in it rather than just a simple `asg_dev_scaledown_schedule_0`, `asg_dev_scaledown_schedule_1` etc?

It looks a little bit strange in the AWS console but technically there's nothing wrong with it

terraform/projects/app-ecs-instances/main.tf
```diff
@@ -216,3 +240,8 @@ output "available_azs" {
   value       = "${data.aws_subnet.subnets_for_prometheus.*.availability_zone}"
   description = "AZs available with running container instances"
 }
+
+output "asg_dev_scaledown_schedules" {
```
idavidmcdonald: I don't know if we particularly need to output this but fine either way

README.md
```diff
@@ -120,6 +120,23 @@ To apply terraform for a list of projects:
 `. ./setup.sh -a list "<1st project>,<2nd project>,<3rd project>"`
 </details>
 
+### Dev EC2 instances lifecycle
+
+By default the EC2 instances on dev stacks will be spun down at the end of the day, typically 6pm UTC on Monday to Fridays, so during BST EC2 instances will be terminated at 7pm. 
+
+If the EC2 dev instance has been terminated, at the start of a day or if you are working after core hours, you can easily spin up your instances using this Makefile command: `make apply-single project=app-ecs-instances` or shell script: `. ./setup.sh -a app-ecs-instances`. 
+
+The schedule will automatically spin down EC2 instances at each hour outside the core working hours of 9am - 6pm UTC, Monday to Fridays. This is to ensure that the instances will be spun down even after they have been restarted.
```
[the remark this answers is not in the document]
kentsanggds: ok

terraform/projects/app-ecs-instances/main.tf
kentsanggds: ok

terraform/projects/app-ecs-instances/main.tf
kentsanggds: Thought it would be useful to see what the schedule would be in case devs have a setting in their `tfvars` but have forgotten about it, as it won't be that visible unless the schedule is changed.

PR #59, opened 2018-07-11
Reduce size of dev EC2 instances

kentsanggds:
## What

Aim of this PR is to reduce the size of the instances as in dev environments it's not necessary to run instances of `m4.xlarge`.

Updated setup.sh to remove the tfvars file when doing a clean of the stack, this is to ensure that when creating new dev stacks it uses the latest variables, previously devs were able to continue using the tfvars without any new vars being added resulting in default values for vars being used. 

In this case rather than using m4.xlarge use t2.small, and to pick up lower prometheus cpu and memoryReservation values for dev stacks.

The default dev stack `tfvars`, which is created by function `create_stack_configs` in the shell script, has been updated with default dev values which can be overridden by devs.

The prometheus task definition was changed in order to use the variables from tfvars for your stack if they exist otherwise default values will be used which are the default values for production and staging.

## Reference 

https://trello.com/c/Gn0fdyOr/447-destroy-our-dev-stacks-overnight-to-reduce-costs

### DIFF
--- a/setup.sh
+++ b/setup.sh
@@ -36,11 +36,14 @@ else
 cat <<EOF >stacks/${ENV}.tfvars
 dev_environment = "true"
 ecs_instance_ssh_keyname = "${ENV}-jumpbox-key"
+ecs_instance_type = "t2.small"
+prom_cpu = "128"
+prom_memoryReservation = "512"
 prometheus_subdomain = "${ENV}"
+prometheis_total = 1
 remote_state_bucket = "${TERRAFORM_BUCKET}"
 stack_name = "${ENV}"
 targets_s3_bucket="$DEFAULT_DEV_TARGETS_S3_BUCKET"
-prometheis_total=1
 additional_tags = {
   "Environment" = "${ENV}"
 }
@@ -59,10 +62,10 @@ create_bucket() {
 }
 
 does_stack_config_exist() {
-        if [ -e "${ROOTPROJ}/stacks/${ENV}.backend" ] ; then
+        if [ -e "${ROOTPROJ}/stacks/${ENV}.tfvars" ] ; then
                 return 0
         else
-                echo "stacks/${ENV}.backend doesn't exist, create the stack config files first"
+                echo "stacks/${ENV}.tfvars doesn't exist, create the stack config files first"
                 return 1
         fi
 }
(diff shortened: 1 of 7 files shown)


idavidmcdonald:
Do you have a rough estimate of potential cost savings for reducing the instance size?

kentsanggds:
Given that:

t2.small costs - $0.025 per Hour
m5.xlarge costs - $0.214 per Hour

Rough estimate of running costs for a week running 8 hours per day per stack x number of dev stacks, and assuming only 1 instance per stack -

0.025 x 8 x 5 = $1 per day x 5 dev instances = $5 per week
0.214 x 8 x 5 = $8.56 per day x 5 dev instances = $42.80 per week

Savings over a year, assuming instances are only running for 48 weeks in a year due to holidays/sickness = 

(42.80 - 5) x 48 = rough saving of $1814.40 over a year

setup.sh
```diff
@@ -44,6 +44,9 @@ prometheis_total=1
 additional_tags = {
   "Environment" = "${ENV}"
 }
+ecs_instance_type = "t2.small"
```
idavidmcdonald: Maybe we could look into ordering these variables in a nicer way (either by grouping similar together or alphabetical)

terraform/projects/app-ecs-services/prometheus-service.tf
```diff
@@ -7,6 +7,21 @@
 *
 */
 
+variable "prom_cpu" {
```
idavidmcdonald: Will need to run the `make docs` command to generate new docs as we have new variables please

setup.sh
```diff
@@ -232,6 +235,12 @@ else
                 if [ $2 ] ; then
                         clean $2
                 else
+                        # For a fresh environment, move tfvars in order to force creation of new stack tfvars
+                        if [ -e "${ROOTPROJ}/stacks/${ENV}.tfvars" ] ; then
+                                mv ${ROOTPROJ}/stacks/${ENV}.tfvars ${ROOTPROJ}/stacks/${ENV}.tfvars.old
```
idavidmcdonald: Any particular reasons for moving it rather than just deleting it? Would people ever need to use/see the old version again?

setup.sh
```diff
@@ -44,6 +44,9 @@ prometheis_total=1
 additional_tags = {
   "Environment" = "${ENV}"
 }
+ecs_instance_type = "t2.small"
+prom_cpu = "128"
```
idavidmcdonald: Any particular reasoning for these values out of interest?

setup.sh
kentsanggds: could order them

terraform/projects/app-ecs-services/prometheus-service.tf
kentsanggds: ok

setup.sh
kentsanggds: some people have settings in their stacks that they would like to keep, I will need to set up my own schedule for downscaling so want to copy those values across

setup.sh
kentsanggds: I haven't tried t2.micro but it might work? The numbers themselves I had to adjust as I downsized the instances to get the task to work

PR #60, opened 2018-07-11
Extract nginx auth proxy into own ECS service

ejrowley:
This commit extracts nginx auth proxy into it it's own service splitting
it away from Prometheus.

Before this commit Prometheus and nginx were co-located in the same
service, nginx communicated with Prometheus using a container link, this
is inconsistent with how nginx communicates with Alert Manager and
because of the basic auth prevents Prometheus from scrapping the other
instances of Prometheus.

New Nignx Service and Task
==========================

This commit defines a new service in `nginx-auth-proxy.tf` which
implements the new `nginx-auth-proxy.json` task.
The nginx auth has been renamed and updated to point to the new
internal dns names of prometheus.

Changes to ALBs
===============

A few resources have been renamed to be more consistent, clearer. We
have been able to remove the listener rules for the external dns. When
Prometheus and Nginx where co-located we needed to route requests to the
appropriate instance of Prometheus, which meant we needed to route the
request to the specific Nginx that Prometheus was paired with. Now we
can route to any Nginx each Nginx knows how to route a request to any
Prometheus. This means we can default to sending all requests to a
single Nginx target group which all three instances of Nginx are
registered to and let Nginx manage the routing to the instances of
Prometheus.

A new set of internal DNS names have been setup and a new set of
listener rules have been set up for routing Prometheus traffic to using
the internal load balancer.

Changes to Prometheus Service and Task
======================================

The old Nginx task has been removed and the new service now targets the
Prometheus container

Changes to Nginx Config
=======================

There is now a default server for responding to health check requests.
There are 3 new servers for responding to the Prometheus checks. If a
request is made to the load balancer directly it should 404 (unless it
is `/health`) at the default server.


terraform/projects/app-ecs-albs/main.tf
```diff
@@ -169,17 +169,15 @@ resource "aws_route53_record" "alerts_alias" {
...
 }
 
-resource "aws_lb_target_group" "nginx_auth_external_endpoint" {
-  count = "${local.infra_network_public_subnets_count}"
-
-  name                 = "${var.stack_name}-ext-tg-${count.index + 1}"
-  port                 = 80
+resource "aws_lb_target_group" "nginx_auth_proxy_external_endpoint" {
+  name                 = "${var.stack_name}-ext-tg"
+  port                 = 8181
```
philandstuff: why port 8181? why not a more standard port like 80 or 8080?  You don't open the port directly on the host (using `hostPort` so I think 80 and 8080 are available to you.  Going with 8181 violates the principle of least surprise to me - it jumps out as being unusual, and it makes me wonder if there's something funny going on.

terraform/projects/app-ecs-services/nginx-auth-proxy.tf
```diff
@@ -0,0 +1,71 @@
...
+  vars {
+    log_group = "${aws_cloudwatch_log_group.task_logs.name}"
+    region    = "${var.aws_region}"
+  }
+}
+
+resource "aws_ecs_task_definition" "nginx_auth_server" {
+  family                = "${var.stack_name}-nginx-auth"
+  container_definitions = "${data.template_file.nginx_auth_container_def.rendered}"
+  task_role_arn         = "${aws_iam_role.alertmanager_task_iam_role.arn}"
```
philandstuff: why are we reusing the alertmanager role for the nginx task here?

terraform/projects/app-ecs-services/templates/nginx-auth-proxy.conf.tpl
```diff
@@ -1,43 +1,72 @@
 server {
-  listen 9090;
+  listen 8181 default_server;
+
+  location /health {
+    return 200 "Static health check";
+  }
```
philandstuff: you said in the PR description that we want to return a 404, but there's nothing here that explicitly sets up that behaviour.  If this is something we care about, do we want to have an explicit 404 here?

philandstuff:
can you tell us a bit about what deploying this would look like and if the deploy would need some sort of coordination or downtime?

terraform/projects/app-ecs-services/nginx-auth-proxy.tf
ejrowley: That is the wrong role, what is interesting is that everything is working regardless. This is because things are co-located and we are relying on the config grabber in the prometheus task. I am reluctant to add a config grabber task to the nginx because we don't need it but perhaps we should for consistency and in case the nginx gets scheduled on a box with out a config grabber.

terraform/projects/app-ecs-albs/main.tf
ejrowley: This was not to clash with the existing endpoint I think, it can be changed to 80 now that we have removed the old listeners and ALB

terraform/projects/app-ecs-services/templates/nginx-auth-proxy.conf.tpl
ejrowley: The default behaviour in this case will be to return 404

terraform/projects/app-ecs-services/nginx-auth-proxy.tf
philandstuff: ew. i'm not massively keen on relying on the config grabber in the prometheus task

terraform/projects/app-ecs-services/nginx-auth-proxy.tf
```diff
@@ -0,0 +1,70 @@
...
+    name      = "nginx-auth-proxy"
+    host_path = "/ecs/config-from-s3/nginx-auth-proxy/conf.d"
+  }
+}
+
+resource "aws_ecs_service" "nginx_auth_service" {
+  name            = "${var.stack_name}-nginx-auth"
+  cluster         = "${var.stack_name}-ecs-monitoring"
+  task_definition = "${aws_ecs_task_definition.nginx_auth_server.arn}"
+  desired_count   = 3
```
kentsanggds: Does this need to be 3, can it be 1 for dev stacks?

PR #65, opened 2018-07-16
Alert if fewer then 2 Prometheus are running

ejrowley:


philandstuff:
it's prometheis 😛 @JonathanHallam smite thee :P

philandstuff:
I think we could combine these checks into a single alert (untested):

```
up{job=~"alertmanager|prometheus"} == 0 and on(job) sum by(job) (up{job=~"alertmanager|prometheus"})
```

Not sure what you think of this, though.

PR #70, opened 2018-08-06
Adds alerts for Prometheus

MuriloDalRi:
Alerts when Prometheus query timings are spiking over capacity and also
when they are very high for an extended period.

### DIFF
--- a/terraform/projects/app-ecs-services/config/alerts/alerts.yml
+++ b/terraform/projects/app-ecs-services/config/alerts/alerts.yml
@@ -36,3 +36,22 @@ groups:
         description: "No file_sd targets were detected.  Is there a problem accessing the targets bucket?"
 
 # this can be improved however maybe this is something we need to focus on in Q2 when working on the support plan
+
+  - alert: Prometheus_Over_Capacity
+    expr: sum by(instance)(rate(prometheus_engine_query_duration_seconds_sum{job="prometheus"}[5m])) > 8
+    for: 10s
+    labels:
+        product: "prometheus"
+        severity: page
+    annotations:
+        summary: "Service is over capacity."
+        description: "The service name is {{ $labels.job }}. The URL experiencing the issue is {{ $labels.instance }}"
+
+  - alert: Prometheus_High_Load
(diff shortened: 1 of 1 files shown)


terraform/projects/app-ecs-services/config/alerts/alerts.yml
```diff
@@ -36,3 +36,22 @@ groups:
         description: "No file_sd targets were detected.  Is there a problem accessing the targets bucket?"
 
 # this can be improved however maybe this is something we need to focus on in Q2 when working on the support plan
+
+  - alert: Prometheus_Over_Capacity
+    expr: sum by(instance)(rate(prometheus_engine_query_duration_seconds_sum{job="prometheus"}[5m])) > 8
+    for: 10s
+    labels:
+        product: "prometheus"
+        severity: page
```
philandstuff: I wrote `type:ticket` for the arbitrarily-chosen tag in https://trello.com/c/Fpir7Ilk/509-enable-prometheus-alerts-to-be-delivered-by-email so can we be consistent with that?

MuriloDalRi: I took that from a blog post and only saw that we're using type later... fixed!

philandstuff: oh well if there's prior art then we can go with `severity` and update the trello card

MuriloDalRi: From this blog post: https://www.robustperception.io/reduce-noise-from-disk-space-alerts

Happy to have a poll at tomorrow's stand up

philandstuff: oh cool. seems there's prior art for `severity: ticket` as well here: https://www.robustperception.io/alerting-on-crash-loops-with-prometheus

and `severity: page` is used in prometheus's own examples: https://prometheus.io/docs/prometheus/latest/configuration/alerting_rules/

so i'm happy to go with it.

PR #72, opened 2018-08-07
Send emails from alertmanager using SES

kentsanggds:
This is the first step to enable integration with zendesk, it uses AWS SES (Simple Email Service) to allow alertmanager to send emails.

In order to test it on dev environment and to enable staging to send emails the recipients email address will need to be verified on the SES AWS console.

On the production environment, production access has been requested which will allow emails to be sent to any recipient and the daily quota has been increased to 200. 

Note - 
- After merging, the alert that always triggers will need to be removed, they have been left in to trigger the alert to send an email.
- Each alertmanager will send an email, so for staging and prod as there are 3 alertmanagers (one for each availability zone), we expect to receive 3 emails per ticket alert.
  - One way of handling this would be to provide a weave mesh for the alertmanagers (https://www.robustperception.io/high-availability-prometheus-alerting-and-notification), unfortunately it appears that ECS only allows a single port to be exposed, and meshing requires an additional port to be open. We should be able to use meshing this when the prometheus and alertmanager run on an instance rather than in ECS.
- as we are using the default alertmanager configuration the default value for `repeat_interval` is 4 hours, this means that the email will be sent out every 4 hours until the alert is resovled.

Reference -

https://trello.com/c/Fpir7Ilk

### DIFF
--- a/terraform/projects/app-ecs-services/alertmanager-service.tf
+++ b/terraform/projects/app-ecs-services/alertmanager-service.tf
@@ -7,6 +7,11 @@
 *
 */
 
+## Locals
+locals {
+  alertmanager_public_fqdns = "${data.terraform_remote_state.app_ecs_albs.alerts_public_record_fqdns}"
+}
+
 ## IAM roles & policies
 
 resource "aws_iam_role" "alertmanager_task_iam_role" {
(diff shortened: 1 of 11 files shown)


terraform/projects/app-ecs-services/alertmanager-service.tf
```diff
@@ -120,13 +128,38 @@ data "pass_password" "pagerduty_service_key" {
 data "template_file" "alertmanager_config_file" {
   template = "${file("templates/alertmanager.tpl")}"
 
+  # Even though we do not wish to send emails from staging, we can still include SMTP set up
```
philandstuff: why don't we want to send emails from staging? can you explain?

terraform/projects/app-ecs-services/config/alerts/alerts.yml
```diff
@@ -55,3 +56,15 @@ groups:
     annotations:
         summary: "Service is approaching capacity."
         description: "The service name is {{ $labels.job }}. The URL experiencing the issue is {{ $labels.instance }}"
+
+# Just for testing, should be removed after testing complete on staging and production
```
philandstuff: this needs to be removed

terraform/projects/app-ecs-services/alertmanager-service.tf
kentsanggds: I think the wording needs to be improved, as we probably only want to target a verified email and don't expect the quota to go above the daily limit of `100`, whereas in production it might. I've already opened up a case to enable production access, it might take a few days for AWS to resolve.

terraform/projects/app-ecs-services/config/alerts/alerts.yml
kentsanggds: Yes, will remove after testing on staging and production, or could otherwise trigger `NoFileSdTargets`

PR #73, opened 2018-08-08
Allow access to prometheus UI to internal users

DavidJeche:
The aim was to permit users access to prometheus is expression browser.
We used IP  white listing in order to contrl access to the enviroment.

This is related to this ticket: https://trello.com/c/xTIIBsPs/313-give-users-access-to-the-prometheus-interface

### DIFF
--- a/terraform/projects/infra-security-groups/main.tf
+++ b/terraform/projects/infra-security-groups/main.tf
@@ -142,7 +142,7 @@ resource "aws_security_group_rule" "monitoring_external_sg_ingress_any_http" {
   from_port         = 80
   protocol          = "tcp"
   security_group_id = "${aws_security_group.monitoring_external_sg.id}"
-  cidr_blocks       = ["0.0.0.0/0"]
+  cidr_blocks       = ["${var.cidr_admin_whitelist}"]
 }
 
 resource "aws_security_group_rule" "monitoring_external_sg_ingress_any_https" {
@@ -151,16 +151,7 @@ resource "aws_security_group_rule" "monitoring_external_sg_ingress_any_https" {
   from_port         = 443
   protocol          = "tcp"
   security_group_id = "${aws_security_group.monitoring_external_sg.id}"
-  cidr_blocks       = ["0.0.0.0/0"]
-}
-
-resource "aws_security_group_rule" "allow_prometheus_access_alertmanager" {
-  type              = "ingress"
-  to_port           = 80
-  from_port         = 80
-  protocol          = "tcp"
-  security_group_id = "${aws_security_group.alertmanager_external_sg.id}"
-  cidr_blocks       = ["0.0.0.0/0"]
+  cidr_blocks       = ["${var.cidr_admin_whitelist}"]
 }
 
 resource "aws_security_group_rule" "allow_prometheus_access_paas_proxy" {
--- a/terraform/projects/infra-security-groups/main.tf
+++ b/terraform/projects/infra-security-groups/main.tf
@@ -118,7 +118,7 @@ resource "aws_security_group" "alertmanager_external_sg" {
   )}"
 }
 
-resource "aws_security_group_rule" "alertmanager_external_sg_ingress_any_http" {
+resource "aws_security_group_rule" "alertmanager_external_sg_ingress_office_http" {
   type              = "ingress"
   to_port           = 80
   from_port         = 80
@@ -136,7 +136,7 @@ resource "aws_security_group_rule" "alertmanager_internal_alb" {
   source_security_group_id = "${aws_security_group.monitoring_internal_sg.id}"
 }
 
-resource "aws_security_group_rule" "monitoring_external_sg_ingress_any_http" {
+resource "aws_security_group_rule" "monitoring_external_sg_ingress_office_http" {
   type              = "ingress"
   to_port           = 80
   from_port         = 80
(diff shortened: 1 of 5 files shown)


terraform/projects/infra-security-groups/main.tf
```diff
@@ -151,16 +151,7 @@ resource "aws_security_group_rule" "monitoring_external_sg_ingress_any_https" {
   from_port         = 443
   protocol          = "tcp"
   security_group_id = "${aws_security_group.monitoring_external_sg.id}"
-  cidr_blocks       = ["0.0.0.0/0"]
-}
-
-resource "aws_security_group_rule" "allow_prometheus_access_alertmanager" {
```
idavidmcdonald: I'm not that good at understanding all these security groups. Could you explain for me why we are able to remove this security group rule please :)

DavidJeche: This is a rule that made through from the way that we previously had prometheus configured. We had more components which required more rules. We now mainly utilise virtual hosts through the LB.

terraform/projects/infra-security-groups/main.tf
```diff
@@ -142,7 +142,7 @@ resource "aws_security_group_rule" "monitoring_external_sg_ingress_any_http" {
   from_port         = 80
   protocol          = "tcp"
   security_group_id = "${aws_security_group.monitoring_external_sg.id}"
-  cidr_blocks       = ["0.0.0.0/0"]
+  cidr_blocks       = ["${var.cidr_admin_whitelist}"]
```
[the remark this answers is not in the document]
DavidJeche: Currently hoping that I caught all any-any for external facing interfaces.

terraform/projects/infra-security-groups/main.tf
DavidJeche: There is a bunch of them so I believe it is worth reviewing and clean up as part of a card.

philandstuff: can you explain clearly and specifically:

 - what the security group did
 - why it is not needed

When you say "we had more components which required more rules" - which components? Which rules?  What was the specific change that made this rule no longer necessary?

DavidJeche: We use to run prometheus and alert manager in fornt of an ALB
with an Nginx server with each service. There where multiple targets
groups which exposed alertmanager to the external facing ALB. This
meant that we needed alot of security group rules in order to
permit access to external facing ALB from the public. We also needed
internal security groups to ensure that communciation was occuring.

We have now added a signle nginx service that enables routing based
host header, including similar rules to ALB's. This means
we now no longer need to many rules.

terraform/projects/infra-security-groups/main.tf
```diff
@@ -105,80 +105,22 @@ resource "aws_security_group" "monitoring_external_sg" {
...
-  )}"
-}
-
-resource "aws_security_group_rule" "alertmanager_external_sg_ingress_any_http" {
-  type              = "ingress"
-  to_port           = 80
-  from_port         = 80
-  protocol          = "tcp"
-  security_group_id = "${aws_security_group.alertmanager_external_sg.id}"
-  cidr_blocks       = ["${var.cidr_admin_whitelist}"]
```
philandstuff: Looking over this rule with @DavidJeche, we can't understand what it was supposed to be doing. It allows http access directly from our office IPs to the internal load balancer.  Why did we ever want this?

If I try to visit the internal ALB (by copypasting it's default DNS record into a browser), it just times out, suggesting that even with this rule, access isn't permitted.  Did it ever work?  Since we always have accessed alertmanager via the nginx auth proxy (not directly), I'm not sure we've ever seen this rule in action.

terraform/projects/infra-security-groups/main.tf
```diff
@@ -105,37 +105,6 @@ resource "aws_security_group" "monitoring_external_sg" {
...
-}
-
-resource "aws_security_group_rule" "alertmanager_internal_alb" {
-  type                     = "ingress"
-  to_port                  = 80
-  from_port                = 80
-  protocol                 = "tcp"
-  security_group_id        = "${aws_security_group.alertmanager_external_sg.id}"
-  source_security_group_id = "${aws_security_group.monitoring_internal_sg.id}"
```
philandstuff: This rule used to allow nginx and prometheus to access the internal ALB (in order to access prometheus and alertmanager for proxying/scraping/alerting).  What is the equivalent now that we've got rid of `alertmanager_external_sg`?

terraform/projects/infra-security-groups/main.tf
```diff
@@ -147,6 +147,17 @@ resource "aws_security_group" "monitoring_internal_sg" {
   )}"
 }
 
+resource "aws_security_group_rule" "permit_internal_any_any" {
+  type      = "ingress"
+  from_port = 0
+  to_port   = 0
```
philandstuff: please can we not be so permissive with ingress rules?  We may need to permit port 80 (and possible 8080? 9090? 9093?) but that doesn't mean we should permit the world.

philandstuff: After looking at this again, I realise that this PR isn't newly introducing this behaviour; this permissive rule is already present in master.  Therefore, it's no reason to block this change.

We should still have a think about how we use our security groups but for now this PR can be merged.

PR #77, opened 2018-08-09
[WIP discussion ]Mesh alertmanagers to prevent alert duplication 

DavidJeche:
The purpose of this PR was to enable alert manager meshing so that the
email are not duplicated when alerts trigger.

This is based on the following card: https://trello.com/c/Fpir7Ilk/509-enable-prometheus-alerts-to-be-delivered-by-email

I had to change the config file to the default name since I had issue's with defining commands.
By not specifying the alertmanager config I was able to just pass the list of mesh target to the task definition.


DavidJeche:
There is some issue with this mainly around the format list section where arguments for the service are defined. This will not work when running a def stack with only one instance.

philandstuff:
Can we have a conversation about your approach here?  It might get the job done but it adds a lot of complexity and I want to be certain that the tradeoff is worthwhile.

DavidJeche:
@philandstuff This sounds god let me know when you would like to have that.

PR #78, opened 2018-08-10
Fix issue with alert manager task definition

DavidJeche:
We needed to fix a problem with the alert manager task definition. This is
because are using a count which creates three task definition but only
one is ever used. We have set this to a static one value now. We intend
to use something in the future that will enable referencing a single
URL.

### DIFF
--- a/terraform/projects/app-ecs-services/alertmanager-service.tf
+++ b/terraform/projects/app-ecs-services/alertmanager-service.tf
@@ -84,7 +84,7 @@ data "template_file" "alertmanager_container_defn" {
 
 resource "aws_ecs_task_definition" "alertmanager_server" {
   count                 = "${length(local.alertmanager_public_fqdns)}"
-  family                = "${var.stack_name}-alertmanager-server"
+  family                = "${var.stack_name}-alertmanager-server-${count.index + 1}"
   container_definitions = "${element(data.template_file.alertmanager_container_defn.*.rendered, count.index)}"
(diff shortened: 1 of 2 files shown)


DavidJeche:
It turns out that it might be possible to configure a single base url, that has multiple IP values. I am currently testing this now to ensure functionality. This does not mean that we get rid of the alerts-1,alerts-2,alerts-3, this just means that we can set one URL on --web.external-url. We can go to it and it route us to anyone of the meshed alert managers and produce results.

We may also be able to add health checks so if an alert manager is down we wont be routed to it.

DavidJeche:
There is no simple way to perform such routing. I have reviewed the listener rules. You can only set one redirect rule.

philandstuff:
I'm not sure what problem you're trying to solve here?

>We needed to fix a problem with the alert manager task definition.

What is the problem?

> This is because are using a count which creates three task definition but only
one is ever used.

It looks like we use all three?

```terraform
 resource "aws_ecs_task_definition" "alertmanager_server" {
   count                 = "${length(local.alertmanager_public_fqdns)}"
   family                = "${var.stack_name}-alertmanager-server"
   container_definitions = "${element(data.template_file.alertmanager_container_defn.*.rendered, count.index)}"
   task_role_arn         = "${aws_iam_role.alertmanager_task_iam_role.arn}"
```

Can you help me understand why you want to make this change? 😕

kentsanggds:
Fix creates task definitions with an index suffix matching the service rather than placing it under the same task definition, also updates the prometheus task definitions so that they match the service index.

PR #79, opened 2018-08-10
manage the shared dev route53 zone better

philandstuff:
Currently, we treat the shared dev.gds-reliability.engineering zone as
a resource that gets `terraform import`-ed into each project
separately.  To ensure that we don't try to remove the shared zone
when running `terraform destroy`, we have some code in setup.sh which
runs `terraform state rm` on the zone to remove it from the statefile,
prior to destroying the project.

This commit changes the shared zone to be managed as a [data source][1]
instead.  This means that our projects can depend on the zone but
don't own it and therefore won't accidentally try to change or destroy
it.  This also means that, longer-term, setup.sh can remove all the
special case code around `terraform import` and `terraform state rm`.

For a transition period, however, the shared zone will still be around
in people's existing state files, so I've edited setup.sh to simply
try to remove the shared zone from state whenever we do any plan,
apply or destroy on infra-networking in a dev environment.  This code
can be removed once every stack has been touched.  I've marked the
code to be removed in clear comments with a timestamp.

(It's not the end of the world if someone does try to destroy the
shared dev zone resource, because terraform will fail to destroy
non-empty route53 zones.  It'll just cause a failure to the user.)

[1]: https://www.terraform.io/docs/providers/aws/d/route53_zone.html

### DIFF
--- a/setup.sh
+++ b/setup.sh
@@ -83,25 +83,13 @@ clean() {
         fi
 }
 
-import_shared_dev_route53 () {
-# Imports into terraform the shared development route 53 zone that had not been set up
-# using Terraform
-
-        # check if the resource exists in the state file
-        SHARED_DEV_SUBDOMAIN=`aws-vault exec ${PROFILE_NAME} -- $TERRAFORMPATH state list $SHARED_DEV_SUBDOMAIN_RESOURCE`
-        if [ "$SHARED_DEV_SUBDOMAIN" = "$SHARED_DEV_SUBDOMAIN_RESOURCE" ] ; then
-                echo "$SHARED_DEV_SUBDOMAIN already imported"
-        else
-                echo "import $SHARED_DEV_SUBDOMAIN_RESOURCE"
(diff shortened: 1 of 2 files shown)


setup.sh
```diff
@@ -83,25 +83,13 @@ clean() {
...
-# Imports into terraform the shared development route 53 zone that had not been set up
-# using Terraform
-
-        # check if the resource exists in the state file
-        SHARED_DEV_SUBDOMAIN=`aws-vault exec ${PROFILE_NAME} -- $TERRAFORMPATH state list $SHARED_DEV_SUBDOMAIN_RESOURCE`
-        if [ "$SHARED_DEV_SUBDOMAIN" = "$SHARED_DEV_SUBDOMAIN_RESOURCE" ] ; then
-                echo "$SHARED_DEV_SUBDOMAIN already imported"
-        else
-                echo "import $SHARED_DEV_SUBDOMAIN_RESOURCE"
-                aws-vault exec ${PROFILE_NAME} -- $TERRAFORMPATH import $SHARED_DEV_SUBDOMAIN_RESOURCE $SHARED_DEV_DNS_ZONE
```
kentsanggds: We can probably also remove the `SHARED_DEV_DNS_ZONE` defined at the top of the file.

PR #82, opened 2018-08-13
Prefix all alert names with team names

idavidmcdonald:
This enables alerts to be easily distinguished between teams
in prometheus and alertmanager

### DIFF
--- a/terraform/projects/app-ecs-services/config/alerts/alerts.yml
+++ b/terraform/projects/app-ecs-services/config/alerts/alerts.yml
@@ -1,7 +1,7 @@
 groups:
-- name: PrometheusCore
+- name: RE_Observe
   rules:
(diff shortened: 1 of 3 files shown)


philandstuff:
I thought I'd do a quick search for best practice discussions and found [this thread](https://groups.google.com/forum/#!topic/prometheus-developers/k3E6bJ2sT1Q).  In summary:

 - Fabian Reinartz suggested that alert names should be like metric names - generic, but we can query for team-specific teams in alert *labels* (not names)
 - Brian Brazil disagreed and suggested that alert names should have some context flattened into them, especially in an example like `LatencyHigh` where each team will have a different definition of "high"

I'd be tempted to leave this open for a day or so to see how everyone feels?  I'm undecided, myself.

PR #85, opened 2018-08-13
Revert "Allow access to prometheus UI to internal users"

MuriloDalRi:
Reverts alphagov/prometheus-aws-configuration-beta#73


philandstuff:
@MuriloDalRi deployed this to staging, but we had forgotten about grafana access to prometheus.  Grafana isn't on the office IPs so deploying to staging caused grafana to lose access to staging prometheus 😞

PR #86, opened 2018-08-13
Remove unused security group

MuriloDalRi:
"Looking over this rule with @DavidJeche, we can't understand what it was supposed to be doing. It allows http access directly from our office IPs to the internal load balancer. Why did we ever want this?

If I try to visit the internal ALB (by copypasting it's default DNS record into a browser), it just times out, suggesting that even with this rule, access isn't permitted. Did it ever work? Since we always have accessed alertmanager via the nginx auth proxy (not directly), I'm not sure we've ever seen this rule in action."

https://github.com/alphagov/prometheus-aws-configuration-beta/pull/73/files#r209182084

### DIFF
--- a/terraform/projects/infra-security-groups/main.tf
+++ b/terraform/projects/infra-security-groups/main.tf
@@ -118,15 +118,6 @@ resource "aws_security_group" "alertmanager_external_sg" {
   )}"
 }
 
-resource "aws_security_group_rule" "alertmanager_external_sg_ingress_any_http" {
-  type              = "ingress"
-  to_port           = 80
-  from_port         = 80
-  protocol          = "tcp"
-  security_group_id = "${aws_security_group.alertmanager_external_sg.id}"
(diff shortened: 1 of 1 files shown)


terraform/projects/infra-security-groups/main.tf
```diff
@@ -118,15 +118,6 @@ resource "aws_security_group" "alertmanager_external_sg" {
   )}"
 }
 
-resource "aws_security_group_rule" "alertmanager_external_sg_ingress_any_http" {
-  type              = "ingress"
-  to_port           = 80
-  from_port         = 80
-  protocol          = "tcp"
-  security_group_id = "${aws_security_group.alertmanager_external_sg.id}"
-  cidr_blocks       = ["${var.cidr_admin_whitelist}"]
```
philandstuff: we can also remove the `cidr_admin_whitelist` variable because it's no longer used

PR #94, opened 2018-08-15
Enable make file to allow access to terraform console

DavidJeche:
Edit the terraform make file to allow us to use terraforms console feature.
This feature allows you to test interpolations and terraform funcations. This
will enable to us to reduce the number of times we run plan and apply. We can
test all our variables against our terraform funcations local before making an apply.

This is one of the thing that slowed down development time for me.
This commit will greatly improve the development process for me and hopefully
for the team also.

Ticket: https://trello.com/c/hLSlzPfT/533-add-console-state-option-to-make-file-to-enable-speedy-development

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -75,5 +75,5 @@ jump: ## Jump onto the first instance on a dev environment
 
 
 .PHONY: create-console
-create_console: ## Create terraform console session, make create-console project=<project name>
+create-console: ## Create terraform console session, make create-console project=<project name>
 	. ./setup.sh -e ${project}
--- a/Makefile
+++ b/Makefile
@@ -72,3 +72,8 @@ taint: ## Taint a resource, make taint project=<project name> resource=<resource
 .PHONY: jump
 jump: ## Jump onto the first instance on a dev environment
 	. ./setup.sh -j
+
+
+.PHONY: create_console
+create_console: ## Create console session
+	. ./setup.sh -e ${project}
--- a/Makefile
+++ b/Makefile
@@ -74,6 +74,6 @@ jump: ## Jump onto the first instance on a dev environment
 	. ./setup.sh -j
 
 
-.PHONY: create_console
-create_console: ## Create console session
+.PHONY: create-console
+create_console: ## Create terraform console session, make create-console project=<project name>
 	. ./setup.sh -e ${project}
--- a/setup.sh
+++ b/setup.sh
@@ -366,7 +366,7 @@ else
                 jumpbox
         ;;
         -e) echo "Starting console session: ${ENV}"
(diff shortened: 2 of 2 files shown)


Makefile
```diff
@@ -72,3 +72,8 @@ taint: ## Taint a resource, make taint project=<project name> resource=<resource
 .PHONY: jump
 jump: ## Jump onto the first instance on a dev environment
 	. ./setup.sh -j
+
+
+.PHONY: create_console
+create_console: ## Create console session
```
idavidmcdonald: Our convention appears to be using `-` instead of `_` for the task name.

Also please could we include an example to our description i.e. `Create console session, make create-console project=<project name>`

setup.sh
```diff
@@ -208,6 +208,17 @@ jumpbox() {
         ssh -At -oStrictHostKeyChecking=no ec2-user@$JUMPBOX ssh -oStrictHostKeyChecking=no ec2-user@$INSTANCE_IP
 }
 
+create_console() {
+        echo $1
+        cd $TERRAFORMPROJ$1
+
+        aws-vault exec ${PROFILE_NAME} -- $TERRAFORMPATH console -var-file=$TERRAFORMTFVARS
+
+        if [ $? != 0 ]; then
```
idavidmcdonald: I'm not that good at bash but do we need this if statement? If the console statement exits, regardless of the exit code we would exit the function anyway as it's the end of it?

DavidJeche: The idea behind this peace of code was to enable the ability t catch unforseeen cases. It is not majorly required but since the console is highly dynamic enviroment we can always forsee possible failure cases.

I am happy to remove if you are happy to continue with that risk. This peace of code takes the exit code from the previously run command and checks it. The main purpose if to prevent the code from hanging.

Makefile
```diff
@@ -72,3 +72,8 @@ taint: ## Taint a resource, make taint project=<project name> resource=<resource
 .PHONY: jump
 jump: ## Jump onto the first instance on a dev environment
 	. ./setup.sh -j
+
+
+.PHONY: create-console
+create_console: ## Create terraform console session, make create-console project=<project name>
```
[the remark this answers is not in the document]
DavidJeche: yep that is true I missed other places where the function is defined.

setup.sh
```diff
@@ -354,6 +365,9 @@ else
         -j) echo "Jump onto instance: ${ENV}"
                 jumpbox
         ;;
+        -e) echo "Starting console session: ${ENV}"
+                create_console
```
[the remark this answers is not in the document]
DavidJeche: yep edited this

PR #96, opened 2018-08-16
Alertmnager mesh direct

DavidJeche:
# Why I am making this change

Allow alertmanager's to mesh with each other in order to
prevent duplicate states. Changes that include creating a new task definition.
This task definition will permit alert managers to mesh and it will also provide a single place
to provide arguments for the task definition. To make it as clean as possible. Please feel
free to change anything you would like.

# What approach I took

Open up a container port 9094. Dynamically add DNS record via start-up script. Change the way arguments are passed to the task definition to enable mesh system reference.

Verify that this can function with one instance i.e dev environments.

### DIFF
--- a/terraform/projects/app-ecs-instances/instance-user-data.tpl
+++ b/terraform/projects/app-ecs-instances/instance-user-data.tpl
@@ -6,6 +6,7 @@ sudo yum install -y aws-cli wget
 REGION="${region}"
 DEVICE="xvdf"
 
+IP_V4_ADDRESS=$(curl http://169.254.169.254/latest/meta-data/local-ipv4)
 echo "[$(date '+%H:%M:%S %d-%m-%Y')] finding current instance ID"
 INSTANCE_ID=$(wget -q -O - http://169.254.169.254/latest/meta-data/instance-id)
 
@@ -75,3 +76,46 @@ echo "ECS_CLUSTER=${cluster_name}" >> /etc/ecs/ecs.config
 yum install -y ecs-init
 start ecs
 service docker start
+
+case "$AZ" in
+        "eu-west-1a")
+            dns_name="mesh-1"
+            ;;
+
+        "eu-west-1b")
+            dns_name="mesh-2"
+            ;;
+
+        "eu-west-1c")
+            dns_name="mesh-3"
+            ;;
+
+         *)
+            echo "no Dns values found"
+            ;;
+
+esac
+
+
+cat <<EOF >/ecs/dns_update.json
+{
+    "Comment": "Update new instance IP address route 53",
+    "Changes": [
+        {
+            "Action": "UPSERT",
+            "ResourceRecordSet": {
+                "Name": "$dns_name.${private_subdomain}",
+                "Type": "A",
+                "TTL": 300,
+                "ResourceRecords": [
+                    {
+                        "Value": "$IP_V4_ADDRESS"
+                    }
+                ]
+            }
+        }
+    ]
+}
+EOF
+
+aws route53 change-resource-record-sets --hosted-zone-id ${dns_zone_id} --change-batch file:///ecs/dns_update.json
--- a/terraform/projects/app-ecs-instances/main.tf
+++ b/terraform/projects/app-ecs-instances/main.tf
@@ -146,6 +146,8 @@ data "template_file" "instance_user_data" {
     cluster_name = "${local.cluster_name}"
     volume_ids   = "${join(" ", aws_ebs_volume.prometheus_ebs_volume.*.id)}"
     region       = "${var.aws_region}"
+    dns_zone_id  = "${data.terraform_remote_state.infra_networking.private_zone_id}"
+    private_subdomain = "${data.terraform_remote_state.infra_networking.private_subdomain}"
   }
 }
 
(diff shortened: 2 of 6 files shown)


terraform/projects/app-ecs-instances/instance-user-data.tpl
```diff
@@ -75,3 +76,46 @@ echo "ECS_CLUSTER=${cluster_name}" >> /etc/ecs/ecs.config
...
+        "eu-west-1b")
+            dns_name="mesh-2"
+            ;;
+
+        "eu-west-1c")
+            dns_name="mesh-3"
+            ;;
+
+         *)
+            echo "no Dns values found"
```
idavidmcdonald: Optional: I know it's super unlikely but I wonder if our script should fail if it ever gets to this point?

terraform/projects/app-ecs-instances/main.tf
```diff
@@ -146,6 +146,8 @@ data "template_file" "instance_user_data" {
     cluster_name = "${local.cluster_name}"
     volume_ids   = "${join(" ", aws_ebs_volume.prometheus_ebs_volume.*.id)}"
     region       = "${var.aws_region}"
+    dns_zone_id  = "${data.terraform_remote_state.infra_networking.private_zone_id}"
```
idavidmcdonald: Optional: Maybe prefixing this so it's `private_dns_zone_id` would be a quick improvement for readability

terraform/projects/infra-networking/main.tf
```diff
@@ -106,6 +108,7 @@ resource "aws_route53_zone" "subdomain" {
 resource "aws_route53_zone" "private" {
   vpc_id = "${module.vpc.vpc_id}"
   name   = "${local.private_subdomain_name}"
+  force_destroy = true
```
idavidmcdonald: Optional: I'm guessing this is added because we will have records added by the API call from the instances which otherwise would stop terraform a terraform destroy? Might be worth adding as a super brief inline comment or in the commit message just so it's clear why this line of code is here

terraform/projects/app-ecs-services/task-definitions/alertmanager-server.json
```diff
@@ -9,12 +9,13 @@
       {
         "containerPort": 9093,
         "hostPort": 9093
+      },
+      {
+        "containerPort": 9094,
+        "hostPort": 9094
       }
     ],
-    "command": [
```
idavidmcdonald: Could we keep this how it was before so that `config.file` and `web.external-url` are still defined in the task definition (it feels nice to keep it in there rather than move it into Terraform which is less related and seems to have all the different commands spread over various parts of the code)? We could then have something like:

```
"command": [
     "--config.file=/etc/alertmanager/alertmanager.yml",
     "--web.external-url=${alertmanager_url}"
     "${peer_commands}"
]
```

terraform/projects/app-ecs-services/alertmanager-service.tf
```diff
@@ -79,6 +98,8 @@ data "template_file" "alertmanager_container_defn" {
     log_group        = "${aws_cloudwatch_log_group.task_logs.name}"
     region           = "${var.aws_region}"
     config_bucket    = "${aws_s3_bucket.config_bucket.id}"
+    alertmanager_url = "--web.external-url=https://${local.alertmanager_public_fqdns[count.index]}"
```
idavidmcdonald: We are defining `alertmanager_url` twice here. If you see my comment below I think we should stick with the first definition and remove this second one.

terraform/projects/app-ecs-services/alertmanager-service.tf
```diff
@@ -7,9 +7,28 @@
...
+variable "prometheis_total" {
+  type        = "string"
+  description = "Desired number of prometheus servers.  Maximum 3."
+  default     = "3"
+}
+
 ## Locals
 locals {
   alertmanager_public_fqdns = "${data.terraform_remote_state.app_ecs_albs.alerts_public_record_fqdns}"
+  alertmanager_mesh = "${formatlist("--cluster.peer=%s.${local.private_subdomain}:9094", var.mesh_urls)}"
```
idavidmcdonald: Optional: If this is using `local.private_subdomain` which is defined a few lines below it might make sense to move `private_subdomain` up the order so it is defined before being used (at least in sense of reading the code).

terraform/projects/app-ecs-services/task-definitions/alertmanager-server.json
DavidJeche: This is generally not a bad idea, I have not considered the concept myself. The idea behind putting the args in terraform is having a single point of reference. This simplifies maintenance a tiny little bit in that you have one point of reference.

I will test this. i suppose my only question is how does 'alertmanager_url' get passed to the config file ?

terraform/projects/app-ecs-instances/instance-user-data.tpl
DavidJeche: hmmm I believe that is a matter of preference. It is an iffy one. :/

should the script fail, I do not know at this stage. these are the last commands, i believe the purpose of this message is to have something in the log file. 

Why would you want the script to fail as I am currently not sure ?

PR #102, opened 2018-08-29
Add routing for registers alerts

MuriloDalRi:
# Why I am making this change

So that Registers alerts become tickets in Zendesk.

# What approach I took

All alerts will go to the email address provided by Registers.

### DIFF
--- a/terraform/projects/app-ecs-services/alertmanager-service.tf
+++ b/terraform/projects/app-ecs-services/alertmanager-service.tf
@@ -154,12 +154,17 @@ data "pass_password" "dgu_pagerduty_service_key" {
   path = "pagerduty/integration-keys/dgu"
 }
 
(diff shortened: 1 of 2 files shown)


terraform/projects/app-ecs-services/alertmanager-service.tf
```diff
@@ -154,6 +154,10 @@ data "pass_password" "dgu_pagerduty_service_key" {
   path = "pagerduty/integration-keys/dgu"
 }
 
+data "pass_password" "registers_zendesk" {
```
idavidmcdonald: Need to add this data into the template file for `alertmanager_config_file`

MuriloDalRi: Well spotted! Have fixed it now.

PR #103, opened 2018-08-29
Add Prometheus module

ejrowley:
This commit deploys prometheus.

There are two modules defined, enclave/network and enclave/prometheus

There are two projects enclave/verify-perf-a/network and
enclave/verify-perf-a/prometheus

The projects instantiate the respective modules with variables for that
particular environment, for example the vpc_id which verify-perf-a specific
is defined in projects/enclave/verify-perf-a/network and passed through to
the network module which can then use the vpc id to build the subnets.

Using this pattern it is easy to deploy prometheus to new verify
environments by creating new enclave projects and specifying the
appropriate variables.

The network module defines two subnets in different availability zones,
whist not strictly necessary (verify only operates out of a single
availability zone) it was felt it was easier to do it now then build with a
single instance and add the second later.

The network also describes the security groups and vpc endpoints. VPC
endpoints are used to reach s3 and the ec2 api. The s3 endpoint is an
example of a "Gateway" endpoint, it makes external IP address routeable
without going over the public internet. The ec2 endpoint is an example of
interface endpoint, again the goal is to access the ec2 api endpoint
without needing to access the public internet. However in this case instead
of using routing rules to reach the endpoint a network interface is added
to the subnet. If you want to reach the service you connect to the local ip
of the network interface instead of the public ip. This caused a problem,
prometheus uses the EC2 API for service discovery, prometheus uses the aws
go sdk which does not allow you to override the ip address used to connect
to the service. The sdk will always try and connect via
ec2.eu-west-2.amazonaws.com (the sdk does let you configure proxies but
that doesn't help in this case) we therefore have to override the address
of ec2.eu-west-2.amazonaws.com in the hosts file. To do this required a
data lookup against the network interfaces and a bit of terraform trickery
to pull out a list of private_ips for the network interfaces. This is a
little hacky so the ec2-endpoint code has been kept in a separate file to
make it easier to remove if decided to adopt a different approach.

The prometheus module defines the prometheus istances, the disks and the
prometheus config. The instances are built using cloud init. The config
file is published to an s3 bucket. The instances poll the s3 bucket and
download the config and restart the prometheus service. Currently the state
of the scripts which manage the process are proof of concept, they work but
with caveats, currently prometheus is restarted every 2 minutes because the
scripts are not able to tell if the config has been modified and there is
no validation of the config. Prometheus reloads config without interruption
to the service and promethues ignores invalid config and continues running
with the old. So whilst the config updating process could be improved the
impact of the existing scripts is minimal.

Currently the configuration assumes it is being ran within a verify
environment this was to speed up development.

A deploy script was created to help with the running of terraform. It is
proof of concept quality too. An example command `./deploy_enclave.sh -e
verify-perf-a -p gds-tech-ops -a apply -s prometheus` will apply the
prometheus terraform the verfiy-perf-a enclave. The script assumes role to
the promethues_deployer role in the verify account, changes directory in to
correct project directory and executes terraform -a is the action so you
can `-a apply` to apply `-a plan` to run a plan and
`-a destroy` to destroy the environment. The apply runs a plan first saving
the plan file to be used in the apply currently the script does not clean
up the plan files.

### DIFF
--- a/terraform/modules/enclave/network/ec2_vpc_endpoint.tf
+++ b/terraform/modules/enclave/network/ec2_vpc_endpoint.tf
@@ -1,0 +1,35 @@
+resource "aws_vpc_endpoint" "ec2" {
+  vpc_id             = "${var.target_vpc}"
+  service_name       = "com.amazonaws.eu-west-2.ec2"
+  vpc_endpoint_type  = "Interface"
+  security_group_ids = ["${aws_security_group.ec2-private-api-interface.id}"]
+  subnet_ids         = ["${aws_subnet.observe.*.id}"]
+}
+
+resource "aws_security_group" "ec2-private-api-interface" {
+  vpc_id = "${var.target_vpc}"
+  name   = "prometheus_to_ec2"
+
+  ingress {
+    protocol    = "tcp"
+    from_port   = 443
+    to_port     = 443
+    cidr_blocks = ["10.0.3.32/27"]
+  }
+
+  tags {
+    Name        = "${var.team}-${var.product}-${var.environment}-ec2-vpc-endpoint-sg"
+    Environment = "${var.environment}"
+    Product     = "${var.product}"
+    Team        = "${var.team}"
+    ManagedBy   = "terraform"
+  }
+}
+
+data "aws_network_interface" "ec2_endpoint_network_interfaces" {
+  count = "${length(aws_vpc_endpoint.ec2.network_interface_ids)}"
+
+  id = "${element(aws_vpc_endpoint.ec2.network_interface_ids, count.index)}"
+
+  depends_on = ["aws_vpc_endpoint.ec2"]
+}
--- a/terraform/modules/enclave/network/main.tf
+++ b/terraform/modules/enclave/network/main.tf
@@ -1,0 +1,185 @@
+resource "aws_subnet" "observe" {
+  count                   = "${length(keys(var.availability_zones))}"
+  availability_zone       = "${element(keys(var.availability_zones), count.index)}"
+  cidr_block              = "${lookup(var.availability_zones, element(keys(var.availability_zones), count.index))}"
+  map_public_ip_on_launch = false
+  vpc_id                  = "${var.target_vpc}"
+
+  tags {
+    Name        = "${var.team}-${var.product}-${var.environment}-subnet-${element(keys(var.availability_zones), count.index)}"
+    Environment = "${var.environment}"
+    Product     = "${var.product}"
+    Team        = "${var.team}"
+    ManagedBy   = "terraform"
+  }
+}
+
+resource "aws_route_table" "observe" {
+  vpc_id = "${var.target_vpc}"
+
+  tags {
+    Name        = "${var.team}-${var.product}-${var.environment}-rt-${element(keys(var.availability_zones), count.index)}"
+    Environment = "${var.environment}"
+    Product     = "${var.product}"
+    Team        = "${var.team}"
+    ManagedBy   = "terraform"
+  }
+}
+
+# A default route via the internet gateway.
+resource "aws_route" "default" {
+  route_table_id         = "${aws_route_table.observe.id}"
+  destination_cidr_block = "0.0.0.0/0"
+  gateway_id             = "${var.internet_gateway_id}"
+}
+
+resource "aws_route" "vpc_peers" {
+  count = "${length(keys(var.vpc_peers))}"
+
+  route_table_id            = "${aws_route_table.observe.id}"
+  destination_cidr_block    = "${element(keys(var.vpc_peers), count.index)}"
+  vpc_peering_connection_id = "${lookup(var.vpc_peers, element(keys(var.vpc_peers), count.index))}"
+}
+
+# Associate route table with observe subnets
+resource "aws_route_table_association" "observe" {
+  count = "${length(keys(var.availability_zones))}"
+
+  subnet_id      = "${element(aws_subnet.observe.*.id, count.index)}"
+  route_table_id = "${aws_route_table.observe.id}"
+}
+
+resource "aws_security_group" "prometheus_instance" {
+  vpc_id = "${var.target_vpc}"
+  name   = "${var.team}-${var.product}-${var.environment}-prometheus-instance"
+
+  tags {
+    Name        = "${var.team}-${var.product}-${var.environment}-prometheus-instance-sg"
+    Environment = "${var.environment}"
+    Product     = "${var.product}"
+    Team        = "${var.team}"
+    ManagedBy   = "terraform"
+  }
+}
+
+resource "aws_security_group_rule" "allow_ssh_from_gds" {
+  type        = "ingress"
+  protocol    = "tcp"
+  from_port   = 22
+  to_port     = 22
+  cidr_blocks = ["${var.cidr_admin_whitelist}"]
+
+  security_group_id = "${aws_security_group.prometheus_instance.id}"
+}
+
+resource "aws_security_group_rule" "allow_access_to_egress_proxies" {
+  type        = "egress"
+  protocol    = "tcp"
+  from_port   = 8080
+  to_port     = 8080
+  cidr_blocks = ["10.0.1.87/32"]
+
+  security_group_id = "${aws_security_group.prometheus_instance.id}"
+}
+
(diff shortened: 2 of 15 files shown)


terraform/modules/enclave/prometheus/prometheus.conf.tpl
```diff
@@ -0,0 +1,24 @@
+global:
+  scrape_interval:     1s
```
idavidmcdonald: Do we want these to be so aggressive? Default is 1m, our PaaS prometheus is every 30 seconds.

terraform/modules/enclave/prometheus/config_bucket.tf
```diff
@@ -0,0 +1,26 @@
...
+resource "aws_s3_bucket_object" "prometheus_config" {
+  bucket  = "${aws_s3_bucket.prometheus_config.id}"
+  key     = "prometheus/prometheus.yml"
+  content = "${data.template_file.prometheus_config_template.rendered}"
+  etag    = "${md5(data.template_file.prometheus_config_template.rendered)}"
+}
+
+resource "aws_s3_bucket" "prometheus_config" {
+  bucket = "gdsobserve-prometheus-config-store"
+  acl    = "private"
```
idavidmcdonald: You may be interested in setting so that all files must be served side encrypted similar how we do for our current stack:
https://github.com/alphagov/prometheus-aws-configuration-beta/blob/master/terraform/projects/app-ecs-services/main.tf#L143

terraform/modules/enclave/network/ec2_vpc_endpoint.tf
```diff
@@ -0,0 +1,35 @@
+resource "aws_vpc_endpoint" "ec2" {
+  vpc_id             = "${var.target_vpc}"
+  service_name       = "com.amazonaws.eu-west-2.ec2"
+  vpc_endpoint_type  = "Interface"
+  security_group_ids = ["${aws_security_group.ec2-interface.id}"]
+  subnet_ids         = ["${aws_subnet.observe.*.id}"]
+}
+
+resource "aws_security_group" "ec2-interface" {
```
idavidmcdonald: Could you explain this one please

terraform/modules/enclave/network/main.tf
```diff
@@ -0,0 +1,185 @@
...
+
+  cidr_blocks = [
+    "10.0.0.0/22",
+    "10.1.0.0/22",
+  ]
+
+  security_group_id = "${aws_security_group.prometheus_instance.id}"
+}
+
+resource "aws_security_group_rule" "node_exporter_from_other_prom" {
```
idavidmcdonald: Could you explain this one please

terraform/modules/enclave/network/main.tf
```diff
@@ -0,0 +1,185 @@
...
+
+resource "aws_vpc_endpoint" "private-s3" {
+  vpc_id          = "${var.target_vpc}"
+  service_name    = "com.amazonaws.eu-west-2.s3"
+  route_table_ids = ["${aws_route_table.observe.id}"]
+
+  policy = "${data.aws_iam_policy_document.s3_vpc_endpoint_policy.json}"
+}
+
+data "aws_iam_policy_document" "s3_vpc_endpoint_policy" {
```
idavidmcdonald: Just wanted to double check we want the VPC endpoint policy to be so permissive?

terraform/modules/enclave/network/outputs.tf
```diff
@@ -0,0 +1,21 @@
...
+  value = [
+    "${aws_security_group.prometheus_instance.id}",
+  ]
+}
+
+output "endpoint_network_interface_ip" {
+  value = "${flatten(data.aws_network_interface.ec2_endpoint_network_interfaces.*.private_ips)}"
+}
+
+output "routing_tabel" {
```
idavidmcdonald: Not a big deal but tabel is misspelt

terraform/modules/enclave/network/ec2_vpc_endpoint.tf
idavidmcdonald: Might be worth calling this `ec2-private-api-interface` or similar

terraform/modules/enclave/prometheus/config_bucket.tf
```diff
@@ -0,0 +1,26 @@
...
+
+resource "aws_s3_bucket_object" "prometheus_config" {
+  bucket  = "${aws_s3_bucket.prometheus_config.id}"
+  key     = "prometheus/prometheus.yml"
+  content = "${data.template_file.prometheus_config_template.rendered}"
+  etag    = "${md5(data.template_file.prometheus_config_template.rendered)}"
+}
+
+resource "aws_s3_bucket" "prometheus_config" {
+  bucket = "gdsobserve-verify-${var.environment}-prometheus-config-store"
```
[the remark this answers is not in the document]
idavidmcdonald: Can change this in another iteration

terraform/modules/enclave/prometheus/main.tf
```diff
@@ -0,0 +1,67 @@
...
+locals {
+  enable_public_ip = "${var.enable_ssh == 1 ? true : false}"
+}
+
+resource "aws_key_pair" "ssh_key" {
+  key_name   = "${var.environment}-prom-key"
+  public_key = "${file("~/.ssh/id_rsa.pub")}"
+}
+
+resource "aws_instance" "prometheus" {
```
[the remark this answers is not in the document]
idavidmcdonald: Speed and not enough value out of them for the cost

terraform/modules/enclave/prometheus/main.tf
```diff
@@ -0,0 +1,67 @@
...
+  }
+}
+
+resource "aws_volume_attachment" "attach-prometheus-disk" {
+  count = "${length(var.subnet_ids)}"
+
+  device_name  = "${var.device_mount_path}"
+  volume_id    = "${element(aws_ebs_volume.promethues-disk.*.id, count.index)}"
+  instance_id  = "${element(aws_instance.prometheus.*.id, count.index)}"
+  skip_destroy = true
```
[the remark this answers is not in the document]
idavidmcdonald: Think we can get rid of this

terraform/modules/enclave/prometheus/cloud.conf
```diff
@@ -0,0 +1,53 @@
...
+    permissions: '0644'
+  - content: ''
+    path: '/etc/apt/sources.list'
+
+runcmd:
+  - "if [ '${aws_ec2_ip}' ]; then echo '${aws_ec2_ip} ec2.eu-west-2.amazonaws.com' >> /etc/hosts; fi"
+  - [ bash, -c, "/root/format_disk.sh"]
+  - [ bash, -c, "mount /dev/xvdh /mnt"]
+  - [bash, -c, "chown -R prometheus /mnt/"]
+  - [systemctl, restart, prometheus]
```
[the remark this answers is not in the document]
DavidJeche: I think this was just to ensure that Prometheus starts successfully as we discussed with Ed. We will just stick to one reboot and start on a new state of the instance. This will do for the time being.

terraform/modules/enclave/prometheus/main.tf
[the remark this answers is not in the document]
DavidJeche: currently going to run a quick test with this what the outcome is but it should be safe to remove. I believe it just creates the attachment in a different way. That most likely preserves the disk.

terraform/modules/enclave/prometheus/prometheus.conf.tpl
ejrowley: We have reverted the scrape interval to 30 seconds

terraform/modules/enclave/prometheus/main.tf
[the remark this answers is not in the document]
ejrowley: we have removed skip destroy

terraform/modules/enclave/network/outputs.tf
ejrowley: Fixed typo

terraform/modules/enclave/network/main.tf
ejrowley: This is a policy around accessing S3, not about accessing individual buckets. Each bucket we control has its own policy. We could restrict access to buckets we control here, but they are created in the prometheus project - we would end up with cyclic dependencies if we tried to wanted to include them here

PR #104, opened 2018-08-30
Add alert for exceeding 80% disk space on registers apps

MatMoore:
# Why I am making this change

Registers currently only have 1 alert. We've made a list of alerts to set up for production, and the first one we agreed to look at is disk usage.

# What approach I took

This is the first alert I've created so I just followed the example. We decided to use 80% as the threshold. The current value for one of the apps I looked at was 30% so I'm not expecting this to be hit in normal circumstances.

I've left the description blank for now - does anyone have any thoughts on what should go here?


terraform/projects/app-ecs-services/config/alerts/registers.yml
```diff
@@ -9,3 +9,22 @@ groups:
     annotations:
         summary: "App {{ $labels.app }} has too many 5xx errors"
         description: "App {{ $labels.app }} has 5xx errors in excess of 25% of total requests"
+
+  - alert: Registers_DiskSpaceExceeded
+    expr: disk_utilization{org="openregister",space="prod"} > 80
```
idavidmcdonald: So here we should actually use `exported_space="prod"` rather than `space="prod"`. We cover this very briefly in https://reliability-engineering.cloudapps.digital/manuals/alerting-with-prometheus-for-paas.html#writing-your-alerting-rule-promql-expression.

The reason for this is the label for `space` actually refers to the space that your `openregister-metric-exporter` app is running in whereas `exported_space` refers to the application the metric refers to. I assume you only want to alert on apps run in your prod space (rather than those for example that are running in staging or sandbox)?

There is a technical reason behind the scenes about why this is happening and I don't think it's perfect so it may be something we try and change in the future.

Unrelated but for further interesting reading on alerting on disk space you may be interested in this article - https://www.robustperception.io/reduce-noise-from-disk-space-alerts. It may be overkill for the moment, especially as your disk usage appears to be very static but it's an interesting thing to know about.

terraform/projects/app-ecs-services/config/alerts/registers.yml
```diff
@@ -9,3 +9,22 @@ groups:
...
+    for: 300s
+    labels:
+        product: "registers"
+    annotations:
+        summary: "App {{ $labels.app }} has exceeded 80% disk space usage for 5 minutes"
+        description: "TODO"
+
+  - alert: Registers_CPUExceeded
+    expr: avg_over_time(cpu{org="openregister",app="beta-multi",space="prod"}[5m])
```
idavidmcdonald: Same comment here as above regarding use of `space` rather than `exported_space`

idavidmcdonald: I'm happy that `avg_over_time` is indeed the moving average we were after based on https://github.com/prometheus/prometheus/issues/383

On another note, if you play around in the prometheus interface and up the range you are averaging over (currently 5m), the graph does indeed become much smoother as we expected. Changing the range to 30s (which is currently how often new metrics are being scraped) gives you the identical graph to it without using `avg_over_time`. It may be worth us playing with this value for your graphs. We could also expand the summary to make it clear this is based on a moving average being above 80% for 5 minutes rather than every single data point being above 80% for the last 5 minutes.

Next step after that is also to look at removing the restriction to a single app but to include all your apps.

PR #107, opened 2018-09-03
Get email address for zendesk alerts from pass

MuriloDalRi:
# Why I am making this change

So we can get Zendesk tickets for Prometheus alerts.

# What approach I took

Using the email address for that queue

### DIFF
--- a/terraform/projects/app-ecs-services/alertmanager-service.tf
+++ b/terraform/projects/app-ecs-services/alertmanager-service.tf
@@ -158,6 +158,10 @@ data "pass_password" "registers_zendesk" {
   path = "receivers/registers/zendesk"
 }
 
+data "pass_password" "observe_zendesk" {
+  path = "receivers/observe/zendesk"
+}
+
 data "template_file" "alertmanager_config_file" {
   template = "${file("templates/alertmanager.tpl")}"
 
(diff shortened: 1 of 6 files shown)


terraform/projects/app-ecs-services/alertmanager-service.tf
```diff
@@ -189,7 +193,7 @@ data "template_file" "alertmanager_dev_config_file" {
     smtp_smarthost         = "email-smtp.${var.aws_region}.amazonaws.com:587"
     smtp_username          = "${aws_iam_access_key.smtp.id}"
     smtp_password          = "${aws_iam_access_key.smtp.ses_smtp_password}"
-    ticket_recipient_email = "${var.ticket_recipient_email}"
+    ticket_recipient_email = "${data.pass_password.observe_zendesk.password}"
```
idavidmcdonald: Not sure if we want to be passing the zendesk details into dev configuration ever to prevent us putting tickets in from our dev stacks?

idavidmcdonald: Check the terraform docs about variables/tfvars as I think the environment variable overwrites but worth testing it out yourself to be confident.

MuriloDalRi: So I tried it and it seems like pass store takes precedence so I removed all references to that variable where it's not needed anymore.

PR #117, opened 2018-09-10
Add scrape of hub policy app

idavidmcdonald:
# Why I am making this change

We want to scrape the Verify policy app which will be implemented with application level metrics.

# What approach I took

Proof of concept so we have gone for the basic solution which is
to use a regex for `policy` on the instance name and hardcode
the port that we know the app is exposed on. This solution is likely
not scalable but we can deal with that later.

Note, I haven't been able to run the infra tests myself as there is currently a bug with two people running the tests at the same which I believe @DavidJeche will be looking at. I believe they would pass though 🤞

### DIFF
--- a/terraform/modules/enclave/network/main.tf
+++ b/terraform/modules/enclave/network/main.tf
@@ -73,11 +73,15 @@ resource "aws_security_group_rule" "allow_ssh_from_gds" {
 }
 
 resource "aws_security_group_rule" "allow_access_to_egress_proxies" {
-  type        = "egress"
-  protocol    = "tcp"
-  from_port   = 8080
-  to_port     = 8080
-  cidr_blocks = ["10.0.1.87/32"]
+  type      = "egress"
+  protocol  = "tcp"
+  from_port = 8080
+  to_port   = 8080
+
+  # Hardcoding these IPs is technical debt: they are IPs associated with
+  # egress-proxy.service.dmz, which is an ELB on a dynamic IP range.
+  # we should find a better way to do this
+  cidr_blocks = ["10.0.1.151/32", "10.0.1.142/32"]
 
   security_group_id = "${aws_security_group.prometheus_instance.id}"
 }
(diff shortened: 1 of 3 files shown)


terraform/modules/enclave/network/main.tf
```diff
@@ -137,6 +137,19 @@ resource "aws_security_group_rule" "node_exporter_from_other_prom" {
...
 }
 
+resource "aws_security_group_rule" "hub_policy" {
+  type      = "egress"
+  protocol  = "tcp"
+  from_port = 50111
+  to_port   = 50111
+
+  cidr_blocks = [
+    "10.1.0.0/22",
```
philandstuff: I think I got the CIDR blocks right here - policy is on high side, and 10.1.0.0/16 is high side? right? @ejrowley

terraform/modules/enclave/prometheus/prometheus.conf.tpl
```diff
@@ -15,7 +15,17 @@ scrape_configs:
     - source_labels: [__meta_ec2_tag_Name]
       regex: '.*\.stubs\..*'
       action: drop
-    - source_labels: [__meta_ec2_private_ip]
-      target_label: private_ip
+    - source_labels: [__meta_ec2_tag_Name]
+      target_label: instance
+
+  - job_name: 'hub_policy'
+    ec2_sd_configs:
```
philandstuff: Missing `metrics_path` here (I snowflaked it and forgot to push to the branch)

idavidmcdonald: Good spot, will fix

terraform/modules/enclave/prometheus/prometheus.conf.tpl
```diff
@@ -22,3 +22,17 @@ scrape_configs:
...
+    - region: "${aws_region}"
+      profile: "${ec2_instance_profile}"
+      port: 50111
+    relabel_configs:
+    - source_labels: [__meta_ec2_tag_Name]
+      regex: 'policy.*'
+      action: keep
+    - source_labels: [__meta_ec2_private_ip]
+      target_label: private_ip
```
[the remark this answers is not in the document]
ejrowley: this can go, it was just an example of capturing a meta label making it public

terraform/modules/enclave/network/main.tf
ejrowley: 10.1.0.0/22 is the high side

PR #118, opened 2018-09-10
Remove filtering of the ip

ejrowley:
The relabel_config was used to remove app network ip addresses by
explicitly keeping non app network private_ips.

In the end this was not required, we have opted to scrape over both the
app network and the mgmt network.

It was thought that service discovery listed all interfaces of an
instance, because some instances have more then one interface then we
would need to pick one, also the instance where attached to two
different networks. For consistency we where going to use the mgmt
network everywhere so we need a way to remove the app network
interfaces.

It transpires that service discovery only lists one interface per
instance and we need to use the app network for scraping targets with
two interfaces so the default behaviour is desired.


philandstuff:
it would be good to have a commit message explaining *why* this was thought necessary, and *why* it actually isn't

PR #126, opened 2018-09-17
Allow Prometheus to be accessed via Verify jumpbox

ejrowley:
# Why I am making this change

Verify use an nginx reverse proxy to expose their tools, graphite,
sensu etc. They have asked us to expose prometheus in the same way.

# What approach I took

The jump box must be able to reach the prometheus instances, we have
created an `allow_web_interface` ingress rule to allow this.

Verify need to know the ip address of the Prometheus instances to be
able to direct traffic to them. Prometheus IP addresses are not hard
coded and can change when instances are rebuilt. We therefore cannot
hard code the IP address in the verify config.
Instead we need to use DNS, verify have a private hosted zone which
we can use to add the addresses of the promethues instances. This code
is encapsulated as a module to enable us to only included it in the
verify project.

### DIFF
--- a/terraform/modules/enclave/verify-dns/main.tf
+++ b/terraform/modules/enclave/verify-dns/main.tf
@@ -1,0 +1,14 @@
+data "aws_route53_zone" "private_hosted_zone" {
+  name   = "${var.hosted_zone_name}"
+  vpc_id = "${var.target_vpc}"
+}
+
+resource "aws_route53_record" "prometheus" {
+  count = "{var.prometheus_private_ip_count}"
+
+  zone_id = "${data.aws_route53_zone.private_hosted_zone.zone_id}"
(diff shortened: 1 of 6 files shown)


kentsanggds:
Code looks good, how can we test that it is working?

terraform/modules/enclave/verify-dns/main.tf
```diff
@@ -0,0 +1,14 @@
+data "aws_route53_zone" "private_hosted_zone" {
+  name   = "${var.hosted_zone_name}"
+  vpc_id = "${var.target_vpc}"
+}
+
+resource "aws_route53_record" "prometheus" {
+  count = "2"
```
idavidmcdonald: Am I right in thinking we could remove the hardcoding of 2 here and use the length of `var.prometheus_private_ips`?

philandstuff:
@ejrowley can you respond to @idavidmcdonald's comments please?

PR #133, opened 2018-09-21
EC2 Prometheus to alert manager ECS

DavidJeche:
# Why I am making this change

In this PR we aim to enable Prometheus to access ECS alert managers which means that it will be able to scrape ECS alert managers and also send alerts when triggered to it. This will allow us to run EC2 based Prometheus and have it integrate with ECS alert managers which will give us to have the ability to fire alerts based on the metrics collected by the EC2 prometheus.

# What approach I took

- We have had to add some outputs that we can use in order to target the ECS services. These were extracted using remote_state and the given as input into Prometheus. 
- Add a cronjob to the Prometheus cloud config to enable the ability to grab alerts from a remote bucket via the sync method. - Upload alerts in the ECS project to the remote bucket in the EC2 stack. 
- Use s3 bucket ID in order to reference bucket this reduces errors. 

Additionally, metrics from node_exporters are now being scraped.

### DIFF
--- a/terraform/modules/enclave/paas-config/prometheus.conf.tpl
+++ b/terraform/modules/enclave/paas-config/prometheus.conf.tpl
@@ -1,6 +1,13 @@
 global:
   scrape_interval: 30s
   evaluation_interval: 30s
+alerting:
+  alertmanagers:
+  - scheme: http
+    static_configs:
+      - targets: ["${alertmanager_dns_names}"]
+rule_files:
+  - "/etc/prometheus/alerts/*"
 scrape_configs:
   - job_name: prometheus
     static_configs:
@@ -11,3 +18,9 @@ scrape_configs:
     file_sd_configs:
       - files: ['/etc/prometheus/targets/*.json']
         refresh_interval: 30s
+  - job_name: alertmanagers
+    static_configs:
+      - targets: ["${alertmanager_dns_names}"]
+  - job_name: prometheus_node
+    static_configs:
+      - targets: ["${prometheus_dns_nodes}"]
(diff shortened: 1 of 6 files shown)


terraform/modules/enclave/paas-config/prometheus.conf.tpl
```diff
@@ -11,3 +18,9 @@ scrape_configs:
     file_sd_configs:
       - files: ['/etc/prometheus/targets/*.json']
         refresh_interval: 30s
+  - job_name: alertmanagers
```
idavidmcdonald: These job doesn't have exactly the same as our ECS file (here alertmanager has been made plural). This means that our alerts we've written for alertmanager will not fire - https://github.com/alphagov/prometheus-aws-configuration-beta/blob/master/terraform/projects/app-ecs-services/config/alerts/observe-alerts.yml

terraform/modules/enclave/prometheus/main.tf
```diff
@@ -92,7 +93,7 @@ resource "aws_security_group_rule" "allow_prometheus_private" {
   security_group_id = "${aws_security_group.allow_prometheus.id}"
   type              = "ingress"
   from_port         = 9090
-  to_port           = 9090
+  to_port           = 9100
```
idavidmcdonald: So I believe this security group allows traffic from EC2 prometheus to the node exporter running on the EC2 instances. We could probably rename the security group rule from `allow_prometheus_private` to something that mentions `node_exporter` too.

We've changed the existing rule rather than adding a new one meaning that the previous rule is no longer needed. Could you explain why please as I wasn't able to figure it out myself. Thanks!

terraform/modules/enclave/paas-config/prometheus.conf.tpl
DavidJeche: Good catch this very important as we would have broken this in the worst way possible

terraform/modules/enclave/prometheus/main.tf
DavidJeche: We have modified the rule so that Prometheus can scrape node exporter on other instances. We have also change the name to make more clear as to the purpose of the rule.

PR #136, opened 2018-09-25
Changes label in alert expression

MuriloDalRi:
We're removing the exported_space labels so these alerts need to be updated.

This needs to be merged after/at the same time as https://github.com/alphagov/cf_app_discovery/pull/16

We should also notify registers that their alerts will stop working during this deployment.


idavidmcdonald:
What is your roll out plan for this? Would be nice to include it in the PR message please as if this was deployed now it would have immediate affects on our users (their alerts could have different behaviour) and we may want to release this in a certain order or at a certain time in relation to the cf_app_discovery PR.

PR #140, opened 2018-09-26
change file_sd targets alert so that it actually works

philandstuff:
The previous version of this alert failed because `count` never
returns 0 -- it returns no series at all.

I fixed this by changing the negative check (prometheus is not
2.1.0+ds) to a positive check (prometheus is 2.4.x)


philandstuff:
yeah I thought about it and left it alone. I think the solution is to fix the version of prometheus we run on EC2

PR #141, opened 2018-09-27
Direct traffic to EC2 prom-3

idavidmcdonald:
# Why I am making this change
https://trello.com/c/d9Oq70D0/562-expose-an-instance-of-the-new-prometheus-on-prom-3

We want to swap traffic from our ECS prom-3 to our EC2 prom-3. 

We will swap traffic for the other two prometheis at a later point.

# What approach I took
We added our first dev enclave project. This was decided during a mob session as the future way forward (at least for the short term) to enable us to develop locally against our ECS stack otherwise we would need to point our enclave at staging and risk breaking things there.

We removed reference for resources from our enclave in the ECS modules but we instead can use ECS resources in our enclave projects. This will mean we have no circular dependancies across our code (which existed before). Also it means eventually code will transition out of the ECS modules into the enclave modules.

### DIFF
--- a/terraform/projects/app-ecs-services/nginx-auth-proxy.tf
+++ b/terraform/projects/app-ecs-services/nginx-auth-proxy.tf
@@ -49,7 +49,7 @@ data "template_file" "nginx-auth-proxy-config-file" {
     alertmanager_3_dns_name = "${data.terraform_remote_state.app_ecs_albs.alerts_private_record_fqdns.2}"
     prometheus_1_dns_name   = "${data.terraform_remote_state.app_ecs_albs.prom_private_record_fqdns.0}"
     prometheus_2_dns_name   = "${data.terraform_remote_state.app_ecs_albs.prom_private_record_fqdns.1}"
-    prometheus_3_dns_name   = "${data.terraform_remote_state.app_ecs_albs.prom_private_record_fqdns.2}"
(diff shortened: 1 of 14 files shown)


terraform/projects/app-ecs-services/nginx-auth-proxy.tf
```diff
@@ -49,7 +49,7 @@ data "template_file" "nginx-auth-proxy-config-file" {
     alertmanager_3_dns_name = "${data.terraform_remote_state.app_ecs_albs.alerts_private_record_fqdns.2}"
     prometheus_1_dns_name   = "${data.terraform_remote_state.app_ecs_albs.prom_private_record_fqdns.0}"
     prometheus_2_dns_name   = "${data.terraform_remote_state.app_ecs_albs.prom_private_record_fqdns.1}"
-    prometheus_3_dns_name   = "${data.terraform_remote_state.app_ecs_albs.prom_private_record_fqdns.2}"
+    prometheus_3_dns_name   = "prom-ec2-3.${data.terraform_remote_state.infra_networking.private_subdomain}:9090"
```
philandstuff: it's a pedantic point, but this thing isn't a dns name because it has a port number.  I'm not sure what it should be. `address`, maybe? That fits prometheus's use of the `__address__` label.

idavidmcdonald: I encourage pedantic comments 👍 I agree. Will look to get this tidied up when we move the other two prometheis over to EC2

PR #147, opened 2018-10-08
[WIP] Add observe alerts tests

kentsanggds:
# Why I am making this change

It's not easy for us or our users to write tests that will trigger correctly with confidence because they are difficult to test, the Prometheus team has just released a feature that allows us to set up the metrics artificially in order to trigger alerts and to test Prom QL expressions and alerts against them.

To demonstrate the alert tests I created RE Observe alerts tests and documented a way of generating the tests and alerts from scratch. As this can be done without too much contact with a live PromQL expression browser, it should be easier to write alerts with confidence faster and helps us and our users understand the metrics and PromQL expressions better.

# What approach I took

The test alerts file was placed in the `alerts-test` folder in the `app-ecs-services` module at the same level of the `config/alerts` folder, and a shell script was created in the `tools` folder in order to allow all tests to be run by default, or to target a teams test file. 

So in order to run the observe tests execute this command - `./tools/test-alerts.sh` from the project root.

Read the docs here - https://github.com/alphagov/prometheus-aws-configuration-beta/blob/dacc0acdf3fbdb7bb8ca3cf0cd34795eaa9bb01f/terraform/modules/app-ecs-services/config/alerts-tests/README.md


terraform/projects/app-ecs-services/config/alerts-tests/README.md
```diff
@@ -0,0 +1,108 @@
+# Writing Tests for Alerts
```
idavidmcdonald: I don't think we want these files duplicated in both `terraform/projects/app-ecs-services...` and `terraform/modules/app-ecs-services`?

terraform/modules/app-ecs-services/config/alerts-tests/README.md
```diff
@@ -0,0 +1,112 @@
+# Writing Tests for Alerts
+
```
idavidmcdonald: Is there any official documentation on the prometheus website to link to about how these tests are created/defined etc?

idavidmcdonald: Ah I think I spotted one below. I think that should go at the top for context.

terraform/projects/app-ecs-services/config/alerts-tests/README.md
kentsanggds: They should be deleted

terraform/modules/app-ecs-services/config/alerts-tests/test-observe-alerts.yml
```diff
@@ -0,0 +1,148 @@
...
+                      job: test
+                      org: test
+                      product: prometheus
+                      space: prod
+                  exp_annotations:
+                      summary: "test target is down"
+                      description: "One of the test targets has been down for 24 hours"
+                      runbook: "https://re-team-manual.cloudapps.digital/prometheus-for-paas-support.html#re-observe-target-down"
+
+      promql_expr_test:
```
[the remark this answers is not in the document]
kentsanggds: The main benefit was to show how we can experiment with the `promql_expr_test`, would probably be more useful to show what can be done with particular types of queries like difference between `rate` and `increase`, how to use `without` and `ignoring` etc

terraform/modules/app-ecs-services/config/alerts-tests/test-observe-alerts.yml
```diff
@@ -0,0 +1,148 @@
...
+
+evaluation_interval: 1m
+
+tests:
+    - interval: 30s
+      input_series:
+          - series: 'prometheus_build_info{job="prometheus",version="2.4.2"}'
+            values: '1+0x14' # 1 1 1 1 1 1 1 1 1 1 1 1 1 1
+
+          - series: 'up{job="alertmanager", instance="localhost:9090"}'
```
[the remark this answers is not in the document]
kentsanggds: I did the bare minimum to get the alerts to fire, but they should probably be updated on an iterative basis to show how alerts will fire in real life.

kentsanggds:
As there's currently an issue with running the tests, this is kind of blocked until the prometheus team can look into it as the tests are failing on Prometheus 2.6 but work on 2.5.

Issue raised here -
https://github.com/prometheus/prometheus/issues/5064

issyl0:
We're not actively working on this (since October), so I'll close it.

PR #148, opened 2018-10-12
Fix dev stacks containing less than 3 prometheis

kentsanggds:
# Why I am making this change

By default dev stacks only have 1 instance, however this causes a problem when getting the fqdns as it was expecting at least 2. This is a quick fix whilst we are in the middle of migrating to EC2 prometheis which will mean that the EC2 prometheus will only be shown when the number of prometheis is 3 not less.

# What approach I took

As we are migrating the EC2 prometheis there is not much point in spending too long on a solution. This solution should get the dev stack working for EC2 prometheis.


philandstuff:
needs `terraform fmt` tho

PR #149, opened 2018-10-12
New prometheus config

philandstuff:
# Why I am making this change

I thought it would be good to have an example config for scraping the log-cache-adapter.

# What approach I took

You run the ./run-prom.sh bash script, and it runs prometheus and a
background job to regularly refresh the oauth bearer token from cloud
foundry.


kentsanggds:
Whats the best way of reviewing this? 

I think I've got it running, not sure what we should be expecting

philandstuff:
I guess have an explore of things you get for the query `{job="log_cache_adapter"}` and see if you get any useful metrics?

kentsanggds:
This is a useful query for working out what kinds of metrics are available - 

`sum by(__name__)({job="log_cache_adapter"})` 

I've added it as part of my PR - https://github.com/alphagov/prometheus-aws-configuration-beta/pull/147, as I think it's a useful way of starting on the journey of writing alerts.

Any ideas as to what the metrics here in the `log_cache_adapter` mean?

philandstuff:
@kentsanggds log-cache-adapter is the thing that will hopefully replace paas-metric-exporter. so there should be cpu, memory, disk usage, requests, that kind of thing.

philandstuff:
@kentsanggds I don't think that matters for the purpose of this PR. This PR is just to show people how they can grab the metrics in the first place

PR #150, opened 2018-10-12
Increase enclave instance size from t2.medium to m4.large

JonathanHallam:
https://trello.com/c/8iVx15jf/630-bump-ec2-prom-instance-size-to-m4large

# Why I am making this change

We're standardising sizing of our prometheus across instances 

# What approach I took

We've only done this for our own Prom instances not the Verify environment for the time being

Paired with @matthewcullum-gds and @idavidmcdonald


philandstuff:
(I was wondering if the answer is that we need another layer of modules to capture a whole environment?)

JonathanHallam:
We had not thought about it, but post comment had the idea that it maybe would require a new layer and some shared variables

PR #152, opened 2018-10-12
Added new grafana down alert.

matthewcullum-gds:
# Why I am making this change

We currently have no alerting for grafana

# What approach I took

This alert measures whether Prometheus has been able to scrape data from
grafana, and alerts after 5 mins of being unable to do so. 

https://trello.com/c/L3TFMNbQ/629-write-alert-for-grafana-down

Query tested in Prometheus

Paired with @JonathanHallam

### DIFF
--- a/terraform/projects/app-ecs-services/config/alerts/observe-alerts.yml
+++ b/terraform/projects/app-ecs-services/config/alerts/observe-alerts.yml
@@ -1,6 +1,16 @@
 groups:
 - name: RE_Observe
   rules:
+  - alert: RE_Observe_Grafana_Down
+    expr: up{job="grafana-paas"} == 0 
+    for: 5m
+    labels:
+        product: "prometheus"
+        severity: "page"
+    annotations:
+        summary: "Prometheus is not able to scrape Grafana"
+        description: "Prometheus has not successfully scraped {{ $labels.job }} in the last 5 minutes. https://grafana-paas.cloudapps.digital/ may be down."
+        runbook: "https://re-team-manual.cloudapps.digital/observe-support.html#re-observe-grafana-down"
   - alert: RE_Observe_AlertManager_Below_Threshold
     expr: up{job="alertmanager"} == 0 and ignoring(instance) sum without(instance) (up{job="alertmanager"}) <= 1
     for: 10s
--- a/terraform/projects/app-ecs-services/templates/alertmanager.tpl
+++ b/terraform/projects/app-ecs-services/templates/alertmanager.tpl
@@ -19,6 +19,10 @@ route:
   - receiver: "registers-zendesk"
     match:
       product: "registers"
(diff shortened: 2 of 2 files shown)


terraform/projects/app-ecs-services/config/alerts/observe-alerts.yml
```diff
@@ -1,6 +1,16 @@
 groups:
 - name: RE_Observe
   rules:
+  - alert: RE_Observe_Grafana_Down
+    expr: up{job="grafana-paas"} == 0 
+    for: 5m
+    labels:
+        product: "grafana"
+        severity: "page"
```
philandstuff: this set of labels doesn't specifically match any receiver in [our alertmanager configuration](https://alerts-1.monitoring.gds-reliability.engineering/#/status).

The default receiver is observe's pagerduty, but it might be good to be more explicit, maybe?

terraform/projects/app-ecs-services/config/alerts/observe-alerts.yml
```diff
@@ -1,6 +1,16 @@
...
+    annotations:
+        summary: "Prometheus is not able to scrape Grafana"
+        description: "The service name is {{ $labels.job }}. The URL experiencing the issue is https://grafana-paas.cloudapps.digital/."
```
philandstuff: please can we not continue copypasting this not-terribly-useful description? :)

terraform/projects/app-ecs-services/config/alerts/observe-alerts.yml
matthewcullum-gds: We edited the alertmanager config to explicitly call pagerduty for this.

idavidmcdonald: I think we still want to use our existing 'prometheus' receiver. 'product' is not the best name but more refers to the team who should be alerted rather than the actual product. If we set it as `prometheus` then our team will be rung. There shouldn't be too much benefit from having an extra `grafana` receiver.

matthewcullum-gds: Amended to use the prometheus receiver.

terraform/projects/app-ecs-services/templates/alertmanager.tpl
```diff
@@ -19,6 +19,10 @@ route:
   - receiver: "registers-zendesk"
     match:
       product: "registers"
+  - receiver: "re-observe-pagerduty"
+    match:
+      product: "grafana"
```
kentsanggds: I think the product should be `prometheus`

matthewcullum-gds: This is now pushed up (we had changed it but it didnt get commited)

PR #159, opened 2018-10-18
Clarify `bundle exec kitchen` in the enclave testing README

issyl0:
```
$ aws-vault exec gds-techops -- kitchen test verify
aws-vault: error: exec: "kitchen": executable file not found in $PATH
```


issyl0:
This doesn't need to be specified, I was just running things in the wrong order. _sigh_

PR #167, opened 2018-10-23
Add -t (target) parameter functionality to deploy script.

matthewcullum-gds:
# Why I am making this change

Although the current deploy_enclave script accepts a -t paramter there
is no code behind it. This changes the script so that it only builds the
thing targeted.

# What approach I took

Script has been modified locally and I have deployed to single instances of my test stack.

### DIFF
--- a/deploy_enclave.sh
+++ b/deploy_enclave.sh
@@ -54,7 +54,7 @@ pushd "terraform/projects/enclave/${ENCLAVE}/${STATE}"
     fi
     if [ "${TERRAFORM_ACTION}" == "apply" ] 
     then
-        aws-vault exec ${PROFILE} -- terraform plan --out "${planfile}"
+        aws-vault exec ${PROFILE} -- terraform plan -target $TARGET --out "${planfile}"
         echo "Do you wish to apply plan?"
         select yn in "yes" "no"; do
             case $yn in


philandstuff:
This change leads me to wonder if we lose as much as we gain by wrapping terraform in a deploy script.  We end up gradually reproducing every terraform feature in the UI of our script.

What would the consequences be if we just deleted deploy.sh and ran terraform directly from the directories?  Then users can use the full feature set of terraform without us needing to duplicate it.

issyl0:
@philandstuff In principle, I agree, but certainly in my experience wrapper scripts are useful to be sure that you're in the right environment (although that doesn't apply so much to us as we `source environment-{staging,production}.sh` for environment specific things), that you're deploying the project you want to deploy (ie, `paas-staging`, vs being in slightly the wrong folder and screwing up). Also, should we ever want to deploy _not_ from our laptops, how do we instrument that?

philandstuff:
@issyl0 

> (although that doesn't apply so much to us as we source environment-{staging,production}.sh for environment specific things)
That only applies to Makefile deploys, not deploy.sh deploys, as far as I'm aware?

I can see your concern about being sure you're in the right directory.  I'd like to see if we can mitigate this risk without isolating ourselves from the tools we use, though.

> Also, should we ever want to deploy not from our laptops?
We can solve that problem if and when it comes up. YAGNI and all that.

issyl0:
@philandstuff 

> That only applies to Makefile deploys, not deploy.sh deploys, as far as I'm aware?

My confusion around two ways of deploying things strikes again!

Go for it then. :-)

PR #171, opened 2018-10-24
Remove ECS Prometheis

issyl0:
# Why I am making this change

We've switched the Prometheis fully over to EC2, so we can remove the ECS config for Prometheus.

For the moment, AlertManager, PaaS Proxy and NGinx are still on ECS, so we can't remove everything.

Paired with @kentsanggds.

https://trello.com/c/DWWeMBr8/662-turn-off-ecs-prometheus-m


issyl0:
I didn't do anything new apart from edit the description slightly and remove a file (we didn't need alert-config.tf as those are the ECS alerts), so I'll merge this with Matthew's review.

PR #173, opened 2018-10-25
Clarify instructions about permissions needed to deploy to Verify

issyl0:
- The comment about extra permissions was useless as it didn't specify
  what extra permissions were needed, or what was prevented. And it
  doesn't seem to apply any more.

### DIFF
--- a/terraform/modules/enclave/README.md
+++ b/terraform/modules/enclave/README.md
@@ -32,19 +32,23 @@ Follow these steps to run infrastructure tests:
 
 ## Deploying
 
-When performing a deployment you need to ensure you have the correct permissions to do so as extra permission are required.
-
 To deploy, run the following script (from the root of this repository):
 
-    ./deploy_enclave.sh -e <enviroment> -p <aws vault profile> -a <terraform method> -s <state> -t <target>
+    ./deploy_enclave.sh -e <environment> -p <aws vault profile> -a <terraform method> -s <state> -t <target>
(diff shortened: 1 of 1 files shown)


terraform/modules/enclave/README.md
```diff
@@ -32,19 +32,17 @@ Follow these steps to run infrastructure tests:
 
 ## Deploying
 
-When performing a deployment you need to ensure you have the correct permissions to do so as extra permission are required.
```
issyl0: According to @kentsanggds, this is relevant to the Verify enclave, not our EC2 enclave. It would be good to specify that, and what extra permissions are needed. I have no idea.

idavidmcdonald: Correct.

Permissions wise you need to have a aws-vault profile as such:

```
[profile verify-perf-a]
mfa_serial = arn:aws:iam::<user_arn>:mfa/<email_address>
role_arn = arn:aws:iam::170611269615:role/prometheus_deployer
```

I believe as well that team members will also need to be added to have access for that role. I don't know if this is something we can do or something we need to ask people with Verify permissions to do for us.

https://reliability-engineering.cloudapps.digital/iaas.html#access-aws-accounts

issyl0:
I've changed the commit message and PR title to be less flippant, and I've clarified the instructions now they relate to things. Thanks, @idavidmcdonald.

PR #174, opened 2018-10-26
Add config_updater task definition

kentsanggds:
# Why I am making this change

This config_udpater task definition was left out which would mean that nginx config updates would not be picked up, which would impact dev stacks as nginx would not have any config files to start with.

This only affects stacks created from scratch so didn't affect staging or production because those stacks already have nginx configs set up.

This does not impact prometheis running in EC2 as it gets the targets configs in the user_data script.

### DIFF
--- a/terraform/projects/app-ecs-services/config-updater.tf
+++ b/terraform/projects/app-ecs-services/config-updater.tf
@@ -1,0 +1,81 @@
+resource "aws_iam_role" "config_task_iam_role" {
+  name = "${var.stack_name}-config-updater-task"
+  path = "/"
+
+  assume_role_policy = <<EOF
+{
+  "Version": "2012-10-17",
+  "Statement": [
+    {
+      "Sid": "",
+      "Effect": "Allow",
+      "Principal": {
+        "Service": "ecs-tasks.amazonaws.com"
+      },
+      "Action": "sts:AssumeRole"
+    }
+  ]
+}
+EOF
+}
+
+data "aws_iam_policy_document" "config_policy_doc" {
+  statement {
+    sid = "GetConfigFiles"
+
+    resources = ["arn:aws:s3:::${aws_s3_bucket.config_bucket.id}/prometheus/*", "arn:aws:s3:::${aws_s3_bucket.config_bucket.id}/alertmanager/*"]
+
+    actions = [
+      "s3:Get*",
+    ]
+  }
+
+  statement {
+    sid = "ListConfigBucket"
+
+    resources = ["arn:aws:s3:::${aws_s3_bucket.config_bucket.id}"]
+
+    actions = [
+      "s3:List*",
+    ]
+  }
+}
+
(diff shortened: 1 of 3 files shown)


terraform/projects/app-ecs-services/paas-proxy.tf
```diff
@@ -64,3 +64,79 @@ resource "aws_s3_bucket_object" "nginx-paas-proxy" {
...
+    config_bucket  = "${aws_s3_bucket.config_bucket.id}"
+  }
+}
+
+resource "aws_ecs_task_definition" "config_updater" {
+  family                = "${var.stack_name}-config-updater"
+  container_definitions = "${data.template_file.config_updater_defn.rendered}"
+  task_role_arn         = "${aws_iam_role.paas_proxy_task_iam_role.arn}"
+
+  volume {
```
issyl0: These are Docker volumes, not EBS volumes, right?

kentsanggds: Not sure what you mean here

issyl0: Don't worry, I answered my own question. :-)

terraform/projects/app-ecs-services/paas-proxy.tf
```diff
@@ -8,31 +8,9 @@ data "template_file" "paas_proxy_container_defn" {
...
-    }
-  ]
-}
-EOF
-}
-
 resource "aws_ecs_task_definition" "paas_proxy" {
   family                = "${var.stack_name}-paas-proxy"
   container_definitions = "${data.template_file.paas_proxy_container_defn.rendered}"
-  task_role_arn         = "${aws_iam_role.paas_proxy_task_iam_role.arn}"
```
idavidmcdonald: Could you explain the removal of this for me please?

kentsanggds: Not needed, the role was only needed to attach a policy to access s3 and paas-proxy doesn't need this

terraform/projects/app-ecs-services/config-updater.tf
```diff
@@ -0,0 +1,81 @@
+resource "aws_iam_role" "config_task_iam_role" {
+  name = "${var.stack_name}-paas-proxy-task"
```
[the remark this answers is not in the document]
kentsanggds: I'll update

terraform/projects/app-ecs-services/config-updater.tf
```diff
@@ -0,0 +1,81 @@
...
+      "Action": "sts:AssumeRole"
+    }
+  ]
+}
+EOF
+}
+
+data "aws_iam_policy_document" "config_policy_doc" {
+  statement {
+    sid = "GetPrometheusFiles"
```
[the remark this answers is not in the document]
kentsanggds: Will update

PR #175, opened 2018-10-26
Add kibana link to alert

matthewcullum-gds:
# Why I am making this change

We are now collecting logs so can link to them


issyl0:
If you want to make the commit history cleaner, instead of having erroneously committed one file and then removed it in a separate commit, you could `git rebase -i master` and squash the two commits together, thus getting rid of the personal stack file like it never existed in history.

matthewcullum-gds:
Thanks I have tidied that, my git fu is a work in progress.

PR #176, opened 2018-10-26
Now all Prometheis are on EC2, use prom-n instead of prom-ec2-n

issyl0:
# Why I am making this change

The Prometheis now run solely in EC2, so we don't need to differentiate them with `ec2` specified in the hostname or the Terraform resource name as there's no conflicting ECS `prom` names.

It would be good to test that this works on others' dev stacks.


issyl0:
I'm not sure this has applied correctly in my dev stack, I still see some evidence of `prom-ec2`.

issyl0:
I hadn't done an `apply` on ECS before an `apply` on enclave. This is working now. :-)

PR #177, opened 2018-10-26
Remove code relating to the Prometheus ECS-managed EBS volumes

issyl0:
- These were 500G volumes attached to ECS-managed EC2 instances. We don't need them
  anymore because we've moved to vanilla EC2.

(Paired with @kentsanggds.)


issyl0:
We need to do some manual steps to Force Detach the volumes in the AWS Console.

PR #180, opened 2018-10-29
Revert "Add back the prom-ec2-n records for now"

issyl0:
Reverts alphagov/prometheus-aws-configuration-beta#179

This makes no sense with what's in my brain about how we do things, so we're rolling it back so we can test it more thoroughly.


issyl0:
Travis isn't updating GitHub, but the builds have passed so I'm merging...

PR #182, opened 2018-10-30
Remove prom-n records from app-ecs-albs

issyl0:
- To prettify the switch of Prometheus from ECS to EC2, we want to
  deprecate the `prom-ec2-n` hosts in favour of `prom-n`. To do this, we
  have to have a multi-step process to ensure that the records are added
  and removed in the correct order with as little downtime as possible.
- This is step 1 of 4.


kentsanggds:
`prom-alias` is being used by `resource "aws_acm_certificate" "monitoring_cert"` and `output "prom_public_record_fqdns"`, so can't remove it until this is resolved

issyl0:
We're making this into another story, so closing this PR for now until we play that other story.

PR #183, opened 2018-10-31
Ship syslog from prometheus for PaaS prod to logit

idavidmcdonald:
# Why I am making this change
https://trello.com/c/thV8SLyW/687-ship-logs-to-logit-aws

# What approach I took
Adds functionality to the prometheus module to send the syslog
of the EC2 instance to logstash using filebeat.

This behaviour is optional as not all stacks will need to send logs
for example dev stacks or the Verify enclave.

We have gone for the most simple configuration for filebeat that
ships syslog but does nothing fancy like adding/editing log lines
or only selecting certain log levels. We made that decision to keep this work
as small as possible but it may be a good idea to look into
changing this in the future. The downside of this is that log lines
don't explicitly tell you which instance of prometheus they are from.
They do however tell you the private IP address which can be
looked up in the EC2 console to give you the corresponding number
prometheus.

We have decided to do only production for the moment so we don't
have to tackle the problem of differentiating between prod
and staging prometheis in our logs (although we would likely solve
this by using a different stack for staging).

You can go see example logs coming from Matt's and my test stacks in the `RE-Observe-Test-Filebeat-Logit` Log stack 

We've also added version control for our logstash file as we are diverting from the default paas logstash configuration. It will still require a manual process to upload any changes to the configuration to Logit.

## Remaining work
- ~~Decide which stack these logs should go to, add the logstash endpoint to the creds store and use it in our terraform code~~ DONE, going into our production log stack
- ~~Decide if we need to explicitly define which number prometheus a log line comes from~~ No, will create a separate card for this
- ~~If these logs get shipped to the same stack as the paas logs, edit the logstash filters to ensure nothing breaks~~ DONE
- Confirm we are happy to not ship staging logs or if not decide where they are shipped to and how they are differentiated from prod logs 

Paired with @matthewcullum-gds

### DIFF
--- a/terraform/modules/enclave/prometheus/filebeat.yml.tpl
+++ b/terraform/modules/enclave/prometheus/filebeat.yml.tpl
@@ -1,0 +1,12 @@
+filebeat.inputs:
+- type: log
+  enabled: true
+  paths:
+    - /var/log/syslog
+
+output.logstash:
+  hosts: ["${logstash_host}"]
+  loadbalance: true
+  ssl.enabled: true
+
+tags: ["prometheus", "${environment}"]
--- a/terraform/modules/enclave/prometheus/main.tf
+++ b/terraform/modules/enclave/prometheus/main.tf
@@ -4,6 +4,7 @@ terraform {
 
 locals {
   enable_public_ip = "${var.enable_ssh == 1 ? true : false}"
+  filebeat_count   = "${var.logstash_host != "" ? 1 : 0}"
 }
 
 resource "aws_key_pair" "ssh_key" {
@@ -70,6 +71,7 @@ data "template_file" "user_data_script" {
     targets_bucket    = "${var.targets_bucket}"
     alerts_bucket     = "${aws_s3_bucket.prometheus_config.id}"
     prom_external_url = "https://${var.prometheus_public_fqdns[count.index]}"
+    logstash_host     = "${var.logstash_host}"
   }
 }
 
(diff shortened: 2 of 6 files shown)


idavidmcdonald:
Found a bug to fix if the `logstash_hosts` isn't provided, for example staging or a dev stack. Will fix up.

terraform/modules/enclave/prometheus/filebeat.yml.tpl
```diff
@@ -0,0 +1,12 @@
...
+  enabled: true
+  paths:
+    - /var/log/syslog
+
+output.logstash:
+  hosts: ["${logstash_hosts}"]
+  loadbalance: true
+  ssl.enabled: true
+
+tags: ["prometheus", "${environment}"]
```
philandstuff: can you update your editor settings to always add newline at end of file please

terraform/modules/enclave/prometheus/main.tf
```diff
@@ -4,6 +4,7 @@ terraform {
 
 locals {
   enable_public_ip = "${var.enable_ssh == 1 ? true : false}"
+  filebeat_count   = "${var.logstash_hosts != "" ? 1 : 0}"
```
philandstuff: urgh terraform

terraform/modules/enclave/prometheus/variables.tf
```diff
@@ -76,3 +76,7 @@ variable "targets_bucket" {
 variable "prometheus_public_fqdns" {
   type = "list"
 }
+
+variable "logstash_hosts" {
```
philandstuff: this variable is called `logstash_hosts` but I'm not sure from reading the code how I would pass more than one host to it.  Can it be either documented (eg with a [variable description](https://www.terraform.io/docs/configuration/variables.html#description-1)) or renamed to `logstash_host` please?

terraform/modules/enclave/prometheus/filebeat.yml.tpl
idavidmcdonald: Damn, don't know why I never did that before... That's gonna save me time

terraform/modules/enclave/prometheus/main.tf
idavidmcdonald: Indeed. Interesting conversation here about the direction Terraform want this to go https://github.com/hashicorp/terraform/issues/18195

idavidmcdonald:
Thanks @philandstuff, think I've addressed all your comments

PR #188, opened 2018-11-09
Add an example alert with comments for what the fields mean

issyl0:
# Why I am making this change

We hypothesise that people would make their alerts more relevant to them if they understood the fields they were filling in.

# What approach I took

731aa0c is an example alert with description above, for every field.
2f7ad65 is an alert with examples for everything in-line.
fe1822d is an alert with examples for the problematic fields (as surfaced in user research) only.

# Rendered versions

[Here](https://github.com/alphagov/prometheus-aws-configuration-beta/blob/731aa0c02ed737f26c3a91c39805744af7c832ca/terraform/projects/app-ecs-services/config/alerts/README.md), [here](https://github.com/alphagov/prometheus-aws-configuration-beta/blob/2f7ad65af066cf611c0859562ad36dd344b2a7bf/terraform/projects/app-ecs-services/config/alerts/README.md) and [here](https://github.com/alphagov/prometheus-aws-configuration-beta/blob/fe1822d021804de2a1f21e62100eb3738ff99a32/terraform/projects/app-ecs-services/config/alerts/README.md) for the three commits.

### DIFF
--- a/terraform/projects/app-ecs-services/config/alerts/README.md
+++ b/terraform/projects/app-ecs-services/config/alerts/README.md
@@ -1,0 +1,45 @@
+# Example Alert
+
+Below is an example alert that you can copy and rewrite to create your
+own alert. [View the RE
+docs](https://reliability-engineering.cloudapps.digital/monitoring-alerts.html#create-and-edit-alerts-using-prometheus)
+for more information on what to consider when writing alerts.
+
+It alerts if the number of 5xx status codes exceeds 25% of total
+requests for 120 seconds (2 minutes) or more.
+
+It is broken down into:
+
+- `alert`: The alert name, in the format `TeamName_Problem`.
+- `expr`: The PromQL query that queries for the data, followed by `>=
+        0.25` defining the threshold of values.
+- `for`: Optional: The alert fires if the query is over threshold for
+         this amount of time.
+- `labels`:
+  - `product`: The team name or product for the team that this alert
+               refers to. For example, "Observe" or "Prometheus".
+- `annotations`:
+  - `summary`: Required: A summary of what the alert shows.
+  - `description`: Required: A more detailed description of what the alert shows.
+  - `dashboard_url`: Optional: A link to your team's dashboard (ie Grafana) to see
+                     trends for the alert.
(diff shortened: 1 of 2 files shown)


idavidmcdonald:
Nice work on doing a few different approaches (with a easy PR message to get to them).

I'm currently leaning towards your first commit (731aa0c is an example alert with description above, for every field.) or if not that then the last one (fe1822d is an alert with examples for the problematic fields (as surfaced in user research) only). Both kind of feel a bit duplicative of our RE docs as we have an example alert there but we might as well try this and test it to see if it is helpful for users.

One suggestion is to put a link to our docs on how to write alerts at the top of every alerts file and also in the README so people head back to our docs for further support too.

matthewcullum-gds:
I find 731aa0c the easiest commit to read. I do tend to check for readme's at the beginning of a file. Failing that the third commit, the inline version was was just to busy and confusing to my eye.

idavidmcdonald:
Fair enough @issyl0. Could a compromise be to add a link to the RE docs on how to write alerts in the README?

philandstuff:
Weak preference for the third commit.

In terms of `labels.app` vs `labels.job`: this is just confusion about the `app` and `job` labels of the underlying metrics matched by the alert.  Probably easier to explain them face-to-face on monday morning, but in brief:

 - `job` is the name of the app that was scraped by prometheus
 - `app` is a label used by paas-metric-exporter to show the name of the app that the metric refers to

So, for example, `cpu{app="jons-way",job="tools-metric-exporter"}` is a metric that records the cpu usage of the `jons-way` app, as reported by the instance of paas-metric-exporter deployed as the app `tools-metric-exporter`.

This is super super confusing for users already when writing promql for dashboards, and the same problem is coming up here when writing alert descriptions.

emilygds:
I also found 731aa0c the easiest to read, too.

One improvement would be to make it clear that this example can be copied, pasted and overwritten to form your first alert. (Otherwise users could go "oh gosh, I have to read all of this.") We want to make clear that this readme gives them what they need. 

It should also reduce the duplication between the RE docs and the readme.

E.g.

"Below is an example alert that you can copy and rewrite to create your alert"

PR #190, opened 2018-11-19
Make AWS Session Manager available on EC2 prometheus

philandstuff:
# Why I am making this change

Currently, you can only ssh to a box if you created it yourself.  If we use AWS Session Manager, we can control shell access using IAM policies, which means we can all access shell without having to manage ssh keys.

We should also be able to get rid of jumpboxes.

# What approach I took

- upgrade to latest ubuntu AMI to get the latest SSM agent
- allow SSM to access the prometheus instances

that's it! 🎉

### DIFF
--- a/terraform/modules/common/ami/main.tf
+++ b/terraform/modules/common/ami/main.tf
@@ -1,5 +1,9 @@
 ## Variables
 
+locals {
+  canonical_account_id = "099720109477"
+}
+
 ## Data sources
 
 data "aws_ami" "ecs_optimized" {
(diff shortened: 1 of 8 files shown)


terraform/modules/common/ami/main.tf
```diff
@@ -1,8 +1,12 @@
 ## Variables
 
+locals {
+  canonical_account_id = "099720109477"
+}
+
 ## Data sources
 
-data "aws_ami" "source" {
+data "aws_ami" "ecs_optimized" {
```
issyl0: Why has this changed to be `ecs_optimized` when the rest of the PR is for EC2 Prometheis?

philandstuff:
Worth mentioning that session manager has no equivalent of port forwarding (that I know of) which is something we should maybe be aware of.  I don't think we need it though.

terraform/modules/common/ami/main.tf
philandstuff: see commit message for 6ec9817c2debe168ac6cb96753fcf4ba38534e0d

issyl0: Oh, I didn't look at the commits individually.

terraform/modules/enclave/README.md
```diff
@@ -52,3 +52,13 @@ To ssh to the instance, with an ssh tunnel to view the web interface (using the
     ssh ubuntu@<ip_from_output> -L 9090:localhost:9090
```
idavidmcdonald: Do we need to remove the ssh instructions here as I assume they are no longer relevant?

philandstuff: i've left ssh in place for the minute while we evaluate session manager, but my hope is that this will go away in future

PR #196, opened 2018-11-21
Refactor infra-networking into a module and per-environment projects

issyl0:
# Why I am making this change

We wanted to refactor the terraform projects, deployable via Makefile commands, into modules and per-environment projects so that we can just use `terraform apply` everywhere and not have to wrap it.

We've started with `infra-networking` as it's the foundation of all of the others.

# What approach I took

We've created a generic `observe-dev` bucket, with `<projectname>-dev` statefile, and the same naming convention for the statefiles in other environments, to avoid blatting the existing ones when we apply this.

There is still a question mark over how we spin up dev stacks, whether people have to change these bucket and tfstate values for themselves, but that's a question for another time.


terraform/modules/infra-networking/main.tf
```diff
@@ -0,0 +1,170 @@
...
+  default     = "monitoring"
+}
+
+# locals
+# --------------------------------------------------------------
+
+locals {
+  default_tags = {
+    Terraform = "true"
+    Project   = "infra-networking"
```
issyl0: TODO: Split this "Project" tag into the project itself.

terraform/projects/infra-networking-dev/main.tf
```diff
@@ -0,0 +1,33 @@
...
+  default     = "ecs-monitoring"
+}
+
+module "infra-networking" {
+  source = "../../modules/infra-networking"
+
+  aws_region      = "eu-west-1"
+  dev_environment = true
+  stack_name      = "module-refactor"
+}
```
philandstuff: we'll need to have `output` stanzas here for outputs to end up in the project statefile, so that they can be consumed by remote_state in other projects. 🍝

issyl0: Which outputs? [All of these](https://github.com/alphagov/prometheus-aws-configuration-beta/pull/196/files#diff-2a21f8981318c66bb6891f0f6b2edc40R105)?

issyl0:
The `terraform state mv` was tedious, but we did it! It involved commands like:

```
aws-vault exec gds-prometheus-staging -- terraform state mv -state=../infra-networking/infra-networking-staging-project-state.json -state-out=./infra-networking-staging-module-state.json aws_eip.nat module.infra-networking.module.vpc.aws_eip.nat
```

terraform/modules/infra-networking/main.tf
```diff
@@ -12,29 +12,11 @@ variable "additional_tags" {
   default     = {}
 }
 
-variable "aws_region" {
-  type        = "string"
-  description = "AWS region"
-  default     = "eu-west-1"
```
philandstuff: did we mean to remove the type/description/default values?

terraform/projects/infra-networking-dev/main.tf
```diff
@@ -0,0 +1,69 @@
+terraform {
+  required_version = "= 0.11.10"
+
+  backend "s3" {
+    bucket = "re-observe-dev"
+    key    = "infra-networking-dev.tfstate"
```
philandstuff: should this be `-modular` like the others, given we hardcode the filename (but allow the bucket name to vary)?

PR #200, opened 2018-11-26
Add the Kubernetes alertmanager to the list of alertmanager targets

issyl0:
# Why I am making this change

We've got an AlertManager in Kubernetes now, on the New Platform. We want to tell our Prometheis to send alerts to it.

# Questions

Will this even work (IP safelisting etc.)? (We tried on one of the production Prometheus instances and it does the HTTP basic auth dance.) Is this the right way of accessing it?

What about staging/production (this is the staging one, but we don't have "normal" alertmanagers in staging)?


terraform/modules/enclave/paas-config/prometheus.conf.tpl
```diff
@@ -5,7 +5,7 @@ alerting:
   alertmanagers:
   - scheme: http
     static_configs:
-      - targets: ["${alertmanager_dns_names}"]
+      - targets: ["${alertmanager_dns_names}", "alertman.cluster.re-managed-observe-staging.aws.ext.govsvc.uk"]
```
philandstuff: this is in a block with `scheme: http` so i don't think it'll work.
you might need a separate block with `scheme: https`?

philandstuff:
> Will this even work (IP safelisting etc.)? (We tried on one of the production Prometheus instances and it does the HTTP basic auth dance.) Is this the right way of accessing it?

Can we add prometheus to the safelist, rather than using basic auth?  I think it should have an elastic IP?

> What about staging/production (this is the staging one, but we don't have "normal" alertmanagers in staging)?

I'm not sure what this question is asking...

PR #201, opened 2018-11-26
restore type/description/defaults

philandstuff:
accidentally removed when we did the project->module refactor, I think?


matthewcullum-gds:
This are the variables that we pass in from the -staging - production projects so I dont believe they are required to be defined in this. Unless I am misunderstanding it.

PR #202, opened 2018-11-26
Reinstate the Project tag for infra-networking resources

issyl0:
- We accidentally removed this in the refactor, for all environments.
- We removed Environment too, but this is now implicit in the Project
  value as there's a project per environment.


issyl0:
`terraform fmt` 👊

PR #204, opened 2018-11-27
Refactor infra-security-groups into a module and per-environment projects

issyl0:
# Why I am making this change

Refactoring the Terraform projects will enable us to run `terraform` commands directly, without wrapper scripts and `source environment_{environment}.sh` to set `.tfvars`.

# What approach I took

Much the same as #196.

The commit history is a little messy, let us know if you'd like us to squash some. It's not as messy as the previous PR though... 😇.


issyl0:
Force pushed to appease `terraform fmt` again and set the correct value for the `project` variable in `infra-security-groups-production`.

terraform/modules/infra-security-groups/main.tf
```diff
@@ -15,52 +14,27 @@ variable "additional_tags" {
   default     = {}
 }
 
-variable "aws_region" {
-  type        = "string"
-  description = "AWS region"
-  default     = "eu-west-1"
-}
+variable "aws_region" {}
```
philandstuff: can we keep the types and descriptions please? (see #201)

terraform/projects/infra-security-groups-production/main.tf
```diff
@@ -0,0 +1,46 @@
...
+}
+
+module "infra-security-groups" {
+  source = "../../modules/infra-security-groups/"
+
+  aws_region          = "${var.aws_region}"
+  stack_name          = "production"
+  remote_state_bucket = "prometheus-production"
+  project             = "${var.project}"
+}
```
philandstuff: missing outputs?

terraform/projects/infra-security-groups-production/main.tf
```diff
@@ -0,0 +1,63 @@
+terraform {
+  required_version = "= 0.11.10"
+
+  backend "s3" {
+    bucket = "prometheus-production"
+    key    = "infra-security-groups-production.tfstate"
```
issyl0: We want to name this `-modular`. We can just rename the statefile in S3?

terraform/modules/infra-security-groups/main.tf
```diff
@@ -17,20 +16,22 @@ variable "additional_tags" {
 
 variable "aws_region" {
   type        = "string"
-  description = "AWS region"
-  default     = "eu-west-1"
+  description = "The AWS region to use."
```
philandstuff: 🤔 this variable is only needed for the remote state data source.  that got me thinking: is it maybe the project's job to wire up the remote state buckets?

Probably not worth doing in this PR but might be worth a discussion.

PR #206, opened 2018-11-27
Disable ssh outside of tests

philandstuff:
Now that we have Session Manager, we don't need ssh access, so we can
close these security group rules and remove these ssh keys.

We still need ssh support in tests for inspec to gain access to check
resources on the machines, so I've left the `enable_ssh` variable as
configurable and left it to `true` in tests.


issyl0:
Ah, re: eee58db, I always wondered how my personal key `~/.ssh/id_rsa` ended up on things rather than `~/.ssh/id_rsa_gds_dev`!

PR #208, opened 2018-11-27
remove the now-dead infra-networking from setup.sh

philandstuff:
to fix makefile-style deployments


issyl0:
I had a branch for this, but thought I'd keep it out and do one `git rm` at the end. But if it's breaking new things then fine.

PR #209, opened 2018-11-27
DO NOT MERGE: fix app-ecs-albs unclean apply problem

philandstuff:
Something has changed, and now app-ecs-albs doesn't apply cleanly.
The problem is that the domain_name in
aws_acm_certificate.monitoring_cert has a trailing dot that apparently
used not to be there.

This commit is a massive hack to just remove the trailing dot with a
regular expression.  But we should probably find a way to avoid
outputting the dot in the first place.


issyl0:
Which environment did you try to deploy this into?

philandstuff:
I ran a `plan` in staging

issyl0:
This is fixed. The trailing dot wasn’t in the code anywhere, nor in our local infra-networking `terraform state mv`ed statefile.

PR #210, opened 2018-11-27
Remove unused target group

philandstuff:
I think this target group is dead.  If you check in the console,
nothing is registered in it.  Nothing in the code references the
prometheus_internal_tg output.  The listener rule only applies to the
internal ALB, but no traffic to prom-1.* is directed to the internal
ALB.


issyl0:
One fewer thing for us to refactor!

PR #212, opened 2018-11-28
[untested] Tighten prom ec2 sg rules

philandstuff:
# Why I am making this change

We had some pretty lax security group rules which allowed all traffic from the whole VPC.  We should be more specific with security group rules.

# What approach I took

Allow the caller of the `prometheus` module to specify a source security group which is allowed to access port 9090.

Warning: this code is untested.  The kitchen tests don't exercise the code path because they allow us to hit prometheus directly.  I'm not really sure how to test it - suggestions, anyone?


philandstuff:
> Rebase please ;)

done

philandstuff:
The latter 🤠

PR #217, opened 2018-11-29
Refactor app-ecs-albs into a module and per-environment projects

issyl0:
# Why I am making this change

Refactoring the Terraform projects will enable us to run terraform commands directly, without wrapper scripts and source environment_{environment}.sh to set .tfvars.

# What approach I took

Much the same as #204. But we're getting faster, with fewer commits. 🤞


issyl0:
Knew we missed at least one thing: need to `git rm -r terraform/projects/app-ecs-albs`

PR #225, opened 2018-12-05
ensure all DGU alerts keep their org and space labels

philandstuff:
To enable #154 to happen, this commit edits the data.gov.uk alerts to
use `without` not `by` clauses, so that we don't unintentionally
remove the `org` and `space` labels everywhere.

The advantage of doing this is that we can then use those labels for
alert routing, and we don't need a separate `product` label at all,
which we can remove at a later date.

The disadvantage is we potentially become a bit noisier about the
labels we're removing - but sometimes explicit is better than
implicit.


philandstuff:
marked as WIP after discussion with @idavidmcdonald - we're going to do some work to get rid of `cf_app_instance` first

philandstuff:
rebased and amended to remove `cf_app_instance` after alphagov/cf_app_discovery#29

PR #227, opened 2018-12-06
keep org and space labels on registers alerts

philandstuff:
like #225 but for the registers alerts


philandstuff:
Marked as WIP after chat with @idavidmcdonald - we're going to get rid of `cf_app_instance` first

philandstuff:
Rebased and modified to remove the `cf_app_instance` label, after alphagov/cf_app_discovery#29

PR #241, opened 2018-12-12
use ec2_sd_configs to find prometheis and node_exporters

philandstuff:
Instead of static_configs

Why? A few reasons:

 - we don't need to hardcode the number of prometheis into the prometheus.yml
 - we can use this in future for discovering alertmanager node_exporters to scrape
 - we can use this in future for discovering alertmanagers to send alerts to

We already had an instance policy that granted `ec2:Describe*` (from the verify-perf-a work, I imagine) so I didn't need to change any IAM stuff to get this to work.

Still to be done:

 - [x] fix the labels so we know which instance is `prom-1`

### DIFF
--- a/terraform/modules/prom-ec2/paas-config/prometheus.conf.tpl
+++ b/terraform/modules/prom-ec2/paas-config/prometheus.conf.tpl
@@ -23,6 +23,11 @@ scrape_configs:
       - source_labels: ['__meta_ec2_tag_Job']
         regex: 'prometheus'
         action: keep
+      - source_labels: ['__meta_ec2_availability_zone']
+        target_label: availability_zone
+      - source_labels: ['__meta_ec2_instance_id']
+        replacement: '$1:9090'
+        target_label: instance
   - job_name: paas-targets
     scheme: http
     proxy_url: 'http://localhost:8080'
@@ -43,3 +48,8 @@ scrape_configs:
       - source_labels: ['__meta_ec2_tag_Job']
         regex: '.+'
         action: keep
+      - source_labels: ['__meta_ec2_availability_zone']
+        target_label: availability_zone
+      - source_labels: ['__meta_ec2_instance_id']
+        replacement: '$1:9100'
+        target_label: instance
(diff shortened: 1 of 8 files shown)


philandstuff:
I've updated this PR.  I've changed the `instance` label to be `instance_id:port`, eg `i-12345abcde:9093`.  I've also added `availability_zone` as a label.

terraform/modules/prom-ec2/paas-config/prometheus.conf.tpl
```diff
@@ -4,26 +4,67 @@ global:
 alerting:
   alertmanagers:
   - scheme: http
-    static_configs:
-      - targets: [${alertmanager_dns_names}]
+    ec2_sd_configs:
+      - region: eu-west-1
```
idavidmcdonald: Optional: Maybe we could be passing the region in as a var?

terraform/modules/prom-ec2/paas-config/prometheus.conf.tpl
```diff
@@ -4,26 +4,67 @@ global:
...
-      - targets: [${prometheus_node_addresses}]
+    ec2_sd_configs:
+      - region: eu-west-1
+        port: 9100
+    relabel_configs:
+      - source_labels: ['__meta_ec2_tag_Environment']
+        regex: '${environment}'
+        action: keep
+      - source_labels: ['__meta_ec2_tag_Job']
+        regex: '.+'
```
idavidmcdonald: Is this regex intentional here? `regex` by default should apply use `.*` But if we only want to get Prometheus nodes (rather than ECS which have no Job label?) then could we more explicit and explicitly ask for instances with  job equal to 'prometheus'?

terraform/modules/prom-ec2/paas-config/prometheus.conf.tpl
philandstuff: YAGNI :)

terraform/modules/prom-ec2/paas-config/prometheus.conf.tpl
philandstuff: The thinking was: match any instance which has a Job label at all (which will therefore exclude ECS).  However, Prometheus doesn't distinguish between an empty label and no label at all, so if you match on `.*` you'll match everything.

I understand about matching `prometheus` explicitly, I guess I was thinking about future-proofing but, er, YAGNI :).  Happy to just match `prometheus`.

PR #247, opened 2018-12-13
point prom-1.* DNS records at new load balancer

philandstuff:
The new load balancer, which skips ECS and points directly at
prometheus, is ready to use.  This PR just points the DNS records for
prometheus at the new load balancer.


idavidmcdonald:
The title of this PR says `prom-1.*` but this does all records, right?

philandstuff:
sorry, yes it's all `prom-*.*` records (although no others)

PR #251, opened 2018-12-18
Add new ALB for ECS alertmanager

philandstuff:
# Why I am making this change

Now that prometheus does not need the nginx auth-proxy ECS service, @idavidmcdonald and I talked about how to remove the last uses of it.  We decided that, as alertmanager doesn't need basic auth, the IP restriction can be implemented by a security group instead of nginx.

In order to achieve that, we need to do a fairly substantial redesign (and simplification) of how alertmanager uses ALBs.

# What approach I took

This PR adds a new ALB for alertmanager in ECS.  The new ALB:

 - is IP-restricted to office IPs at the security group level
 - skips the auth-proxy nginx and goes straight to the backend
 - uses an ACM certificate purely for alertmanager (no prom-* [SANs](https://en.wikipedia.org/wiki/Subject_Alternative_Name))

The new ALB is not in use yet.  To switch over, a future PR needs to:

 - change the `alerts-*` DNS records to point at the new ALB, not the old
 - change the alertmanager ECS service to point at the new target group, not the old

This will cause a period of downtime, but only for office users of the alertmanager UI - not for actual alert delivery.  As this is not a critical use case, I think this downtime would be acceptable.

Once that is done, we can then delete lots of stuff:

 - all the nginx auth-proxy ECS services
 - the old monitoring_external and monitoring_internal load balancers, and associated:
   - listeners, listener rules, target groups in app-ecs-albs
   - security groups in infra-security-groups

We may also be able to remove any internal DNS names, because (after #241 is merged) auth-proxy is the last thing actually using those names, I think.

### DIFF
--- a/terraform/modules/app-ecs-albs/main.tf
+++ b/terraform/modules/app-ecs-albs/main.tf
@@ -53,6 +53,21 @@ variable "alertmanager_count" {
   default     = "3"
 }
 
+variable "alerts_allowed_cidrs" {
+  type        = "list"
+  description = "List of CIDRs which are able to access the alertmanager ALB, default are GDS ips"
+
+  default = [
+    "213.86.153.212/32",
+    "213.86.153.213/32",
+    "213.86.153.214/32",
+    "213.86.153.235/32",
+    "213.86.153.236/32",
+    "213.86.153.237/32",
+    "85.133.67.244/32",
+  ]
+}
+
 # locals
 # --------------------------------------------------------------
 
@@ -136,40 +151,6 @@ resource "aws_lb" "monitoring_internal_alb" {
     map("Stackname", "${var.stack_name}"),
     map("Name", "${var.stack_name}-monitoring-internal-${count.index + 1}")
   )}"
-}
-
-# AWS should manage the certificate renewal automatically
-# https://docs.aws.amazon.com/acm/latest/userguide/managed-renewal.html
-# If this fails, AWS will email associated with the AWS account
-resource "aws_acm_certificate" "alertmanager_cert" {
-  domain_name       = "alerts.${local.subdomain}"
-  validation_method = "DNS"
-
-  subject_alternative_names = ["${aws_route53_record.alerts_alias.*.fqdn}"]
-
-  lifecycle {
-    # We can't destroy a certificate that's in use, and we can't stop
-    # using it until the new one is ready.  Hence
-    # create_before_destroy here.
-    create_before_destroy = true
-  }
-}
-
-resource "aws_route53_record" "alertmanager_cert_validation" {
-  # Count matches the domain_name plus each `subject_alternative_domain`
-  count = "${1 + local.alerts_records_count}"
-
-  name       = "${lookup(aws_acm_certificate.alertmanager_cert.domain_validation_options[count.index], "resource_record_name")}"
-  type       = "${lookup(aws_acm_certificate.alertmanager_cert.domain_validation_options[count.index], "resource_record_type")}"
-  zone_id    = "${var.zone_id}"
-  records    = ["${lookup(aws_acm_certificate.alertmanager_cert.domain_validation_options[count.index], "resource_record_value")}"]
-  ttl        = 60
-  depends_on = ["aws_acm_certificate.alertmanager_cert"]
-}
-
-resource "aws_acm_certificate_validation" "alertmanager_cert" {
-  certificate_arn         = "${aws_acm_certificate.alertmanager_cert.arn}"
-  validation_record_fqdns = ["${aws_route53_record.alertmanager_cert_validation.*.fqdn}"]
 }
 
 resource "aws_route53_record" "alerts_alias" {
@@ -364,6 +345,9 @@ resource "aws_security_group" "prometheus_alb" {
   vpc_id      = "${local.vpc_id}"
 }
 
+# Prometheus is fronted by an nginx which controls access to
+# either approved IP addresses, or users with basic auth creds
+# so we need to allow all IPs here
 resource "aws_security_group_rule" "prom_allow_http" {
   security_group_id = "${aws_security_group.prometheus_alb.id}"
   type              = "ingress"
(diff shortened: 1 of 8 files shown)


terraform/modules/app-ecs-albs/main.tf
```diff
@@ -279,14 +287,97 @@ resource "aws_route53_record" "alerts_private_record" {
...
+resource "aws_security_group" "prometheus_alb" {
+  name        = "${var.stack_name}-prometheus-alb"
+  description = "Access to prometheus ALB (${var.stack_name})"
+  vpc_id      = "${local.vpc_id}"
+}
+
+# Prometheus is fronted by an nginx which controls access to
+# either approved IP addresses, or users with basic auth creds
+# so we need to allow all IPs here
+resource "aws_security_group_rule" "prom_allow_http" {
```
idavidmcdonald: Optional: If I understand this correctly, it might be good to change the names of these two rules to
`prom_alb_allow_http` and `prom_alb_allow_https` so it's clear this is about allowing public traffic to the ALB but not the prometheus EC2 instance itself. The comment above this rule is also useful but could be more specific around what 'here' is (which I think is the ALB).

terraform/modules/app-ecs-albs/main.tf
```diff
@@ -279,14 +287,97 @@ resource "aws_route53_record" "alerts_private_record" {
...
+}
+
 resource "aws_lb" "prometheus_alb" {
   name               = "${var.stack_name}-prometheus-alb"
   internal           = false
   load_balancer_type = "application"
-  security_groups    = ["${data.terraform_remote_state.infra_security_groups.monitoring_external_sg_id}"]
+
+  security_groups = [
+    "${data.terraform_remote_state.infra_security_groups.monitoring_external_sg_id}",
```
idavidmcdonald: We still need the monitoring_external_sg to be attached to the ALB?

terraform/modules/app-ecs-albs/main.tf
```diff
@@ -352,12 +443,189 @@ resource "aws_lb_listener_rule" "prom_listener_https" {
...
+      "alerts-${count.index + 1}.*",
+    ]
+  }
+}
+
+resource "aws_lb_target_group" "alertmanager_tg" {
+  count = "${var.alertmanager_count}"
+
+  name                 = "${var.stack_name}-alerts-${count.index +1}-tg"
+  port                 = 80
```
idavidmcdonald: If I understand this correctly, our target group will send traffic to our alertmanager on port 80? But at the moment we likely want this to go to port 9093 which our ECS instances expose for alertmanager? I might have misunderstood something here though

terraform/projects/app-ecs-albs-dev/main.tf
```diff
@@ -39,13 +39,25 @@ variable "project" {
...
+
 module "app-ecs-albs" {
   source = "../../modules/app-ecs-albs/"
 
   aws_region          = "${var.aws_region}"
   stack_name          = "${var.stack_name}"
   remote_state_bucket = "${var.remote_state_bucket}"
   project             = "${var.project}"
+  zone_id             = "${data.terraform_remote_state.infra_networking.public_zone_id}"
+  subnets             = "${data.terraform_remote_state.infra_networking.public_subnets}"
```
idavidmcdonald: Optional: Wonder if we should always be explicit that these are `public_subnets` rather than just any `subnets`?

terraform/modules/app-ecs-albs/main.tf
```diff
@@ -279,14 +287,97 @@ resource "aws_route53_record" "alerts_private_record" {
...
 }
 
+######################################################################
+# ----- prometheus public ALB -------
+######################################################################
+
+# AWS should manage the certificate renewal automatically
+# https://docs.aws.amazon.com/acm/latest/userguide/managed-renewal.html
+# If this fails, AWS will email associated with the AWS account
+resource "aws_acm_certificate" "prometheus_cert" {
```
idavidmcdonald: So I see that you have introduced a different cert for prom and a different cert for alertmanager. Could you explain your rough reasoning please why this is beneficial as I don't have any context of why we want to do this?

terraform/modules/app-ecs-albs/main.tf
idavidmcdonald: Ah I see your commit message saying this can later be removed 👍

terraform/modules/app-ecs-albs/main.tf
philandstuff: That’s a good question. My thinking is a bit vague but the following factors contributed to the decision:

- principle of least authority: the alermanager ALB doesn’t need a prom-1 cert, so it shouldn’t have one. 
- a view towards extracting common functionality later: the alert manager ALB and the Prometheus ALB have very similar functionality. I’ve been grouping the resources for each together in the app-ecs-albs module. The old certificate was awkward in that it was one of very few pieces of shared functionality which would make a future extracting of a common “ALB” module more difficult.

terraform/modules/app-ecs-albs/main.tf
philandstuff: Happy with this change

terraform/modules/app-ecs-albs/main.tf
philandstuff: 🤔 you know, I’m not at all sure though I think you are probably right

terraform/projects/app-ecs-albs-dev/main.tf
philandstuff: I deliberately chopped `public_` off these names because I think that if app-ecs-albs is offering it as an input, then app-ecs-albs shouldn’t care what kind of subnet it is.

philandstuff: I deliberately chopped `public_` off these names because I think that if app-ecs-albs is offering it as an input, then app-ecs-albs shouldn’t care what kind of subnet it is.

terraform/modules/app-ecs-albs/main.tf
idavidmcdonald: Thanks, good to go with this. Will add some of this context to the commit message

terraform/modules/app-ecs-albs/main.tf
idavidmcdonald: I think I misread your commit message and got confused about which security group it was referring to. I think we do need to remove this now (or when we do the traffic swap over) though as it allows all public traffic to alertmanager. I can see no benefit of keeping it at the moment however so will remove it now.

idavidmcdonald: No I got confused again, this is fine for the moment as this is the prometheus ALB whilst I was thinking of the alertmanager ALB.

PR #258, opened 2018-12-31
Dead mans switch

matthewcullum-gds:
# Why I am making this change

We want to be confident that an alert that fires in Prometheus will be sent from Alertmanager by using a dead mans switch.

# What approach I took

Autom8 setup a Cronitor end point for us to hit. This end point alerts by emailing a local email address should it not get an alert every two minutes. (Once we are happy this is all working as expected the Cronitor config will be changed to raise a page for Production or a ticket for Staging). In dev we setup a constantly firing alert to reach this end point, with the configuration confirmed in dev this was deleted and then replicated for a production and staging end point.


terraform/projects/app-ecs-services-staging/main.tf
```diff
@@ -23,6 +23,12 @@ variable "stack_name" {
   default     = "staging"
 }
 
+variable "dead_mans_switch_cronitor" {
+  type        = "string"
+  description = "Cronitor URL we need to constantly hit for the dead mans switch."
+  default     = "https://cronitor.link/OExIot/run"
```
tlwr: We should treat these URLs as private (but not secret)

PR #264, opened 2019-01-09
Implemented new security groups with sane naming scheme.

matthewcullum-gds:
Added new security groups and added to the resources, removal of the old
SGs will happen in a later PR.
Currently egress is allowed to any this may be tightened at a later
date.

Pair: matthewcullum-gds idavidmcdonald

# Why I am making this change

The old security groups were named badly and did not reflect what they did

# What approach I took

Has been applied to the dev stack to ensure that we can deploy with no down time and that we have retained all functionality.

### DIFF
--- a/terraform/modules/infra-security-groups/main.tf
+++ b/terraform/modules/infra-security-groups/main.tf
@@ -28,6 +28,21 @@ variable "project" {
   description = "Which project, in which environment, we're running"
 }
 
+variable "allowed_cidrs" {
+  type        = "list"
+  description = "List of CIDRs which are able to access the alertmanager ALB, default are GDS ips"
+
+  default = [
+    "213.86.153.212/32",
+    "213.86.153.213/32",
+    "213.86.153.214/32",
+    "213.86.153.235/32",
+    "213.86.153.236/32",
+    "213.86.153.237/32",
+    "85.133.67.244/32",
+  ]
+}
+
 # locals
 # --------------------------------------------------------------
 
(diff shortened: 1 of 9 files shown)


terraform/modules/infra-security-groups/main.tf
```diff
@@ -53,7 +68,184 @@ data "terraform_remote_state" "infra_networking" {
...
+  from_port                = 9090
+  protocol                 = "tcp"
+  source_security_group_id = "${aws_security_group.prometheus_ec2.id}"
+}
+
+resource "aws_security_group_rule" "ingress_from_prometheus_to_prometheus_node_exporter" {
+  security_group_id        = "${aws_security_group.prometheus_ec2.id}"
+  type                     = "ingress"
+  to_port                  = 9100
+  from_port                = 9090
```
philandstuff: I don't understand why this line isn't also `9100`?

`from_port` and `to_port` define a port range, so this rule opens all ports from 9090 to 9100 inclusive, from prometheus_ec2 to itself.

terraform/modules/infra-security-groups/main.tf
```diff
@@ -53,7 +68,184 @@ data "terraform_remote_state" "infra_networking" {
...
+  source_security_group_id = "${aws_security_group.prometheus_ec2.id}"
+}
+
+# This rule allows all egress out of prometheus_ec2. This is for the following purposes:
+# - downloading packages from package repos
+# - calling AWS APIs such as SSM, S3 and EC2
+# - scraping alertmanager on port 9093
+# - sending alerts to alertmanager on port 9093
+# - scraping external targets that run on the PaaS
+# - scraping itself on port 9090
```
philandstuff:
```suggestion
# - scraping itself and other prometheis on port 9090
```

terraform/modules/infra-security-groups/main.tf
```diff
@@ -53,7 +68,184 @@ data "terraform_remote_state" "infra_networking" {
...
+  from_port                = 9093
+  to_port                  = 9093
+  protocol                 = "tcp"
+  source_security_group_id = "${aws_security_group.alertmanager_alb.id}"
+}
+
+resource "aws_security_group_rule" "ingress_from_prometheus_ec2_to_alertmanager_ec2" {
+  security_group_id        = "${aws_security_group.alertmanager_ec2.id}"
+  type                     = "ingress"
+  from_port                = 9090
```
philandstuff: ditto, I think this from_port should be `9093` too

matthewcullum-gds: Explained understood and corrected thanks :-)

terraform/modules/infra-security-groups/main.tf
matthewcullum-gds: Added.

terraform/modules/infra-security-groups/main.tf
matthewcullum-gds: Explained understood and corrected thanks :-)

PR #267, opened 2019-01-11
remove external alertmanager from staging

philandstuff:
this doesn't seem to be working now and the old prometheus k8s cluster
is not being maintained


matthewcullum-gds:
Is this not just part of Chris in build and run tearing down environments and rebuilding them today?

philandstuff:
i don't know, you tell me

philandstuff:
@matthewcullum-gds is right 👍

PR #271, opened 2019-02-06
Use node_creation_time with predict_linear

matthewcullum-gds:
Use the new node_creation_time metric with the predict_linear function
to reduce the number of false alerts. Now the alert will only fire if
the node has been created for the entire evaluation time. This prevents
newly spun up instances from falsely alerting.

Single: matthewcullum-gds


issyl0:
I reran Travis and it was green, so this looks good to me if the query works!

PR #272, opened 2019-02-25
open prod alertmanager to verify egress IPs

philandstuff:
this will allow verify's prometheis to use these alertmanagers


philandstuff:
also, `terraform fmt` is weird with comments which is why the line spacing is whack

PR #280, opened 2019-03-07
Now we've switched over, send Verify Hub ECS alerts to #verify-2ndline

issyl0:


issyl0:
There was a stray `.ruby-version`, hence the force push.

PR #281, opened 2019-03-08
Send Verify P1 alerts to Slack as well as PagerDuty

issyl0:
# Why I am making this change

Previously, all alerts *except* P1s went to Slack, P1s only went to
PagerDuty. For visibility for people not on-call or following along from
home, we want P1s to hit Slack too.

# What approach I took

This adds the Slack config to the "verify-p1" receiver that sends to
PagerDuty, as per
https://www.robustperception.io/sending-alert-notifications-to-multiple-destinations,
with some YAML anchors.

This also changes the Slack receiver from `autom8` to `verify-2ndline`
to reflect what the destination Slack channel actually is now.

### DIFF
--- a/terraform/modules/app-ecs-services/templates/alertmanager.tpl
+++ b/terraform/modules/app-ecs-services/templates/alertmanager.tpl
@@ -92,7 +92,7 @@ receivers:
   - send_resolved: false
     url: "${verify_joint_cronitor}"
 - name: "verify-2ndline-slack"
-  slack_configs:
+  slack_configs: &verify-2ndline-slack-configs
   - send_resolved: true
     channel: '#verify-2ndline'
     icon_emoji: ':verify-shield:'
@@ -100,3 +100,5 @@ receivers:
 - name: "verify-p1"
   pagerduty_configs:
     - service_key: "${verify_p1_pagerduty_key}"
+  slack_configs: *verify-2ndline-slack-configs
+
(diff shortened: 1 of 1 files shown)


terraform/modules/app-ecs-services/templates/alertmanager.tpl
```diff
@@ -91,12 +91,14 @@ receivers:
...
+- name: "verify-2ndline-slack"
+  slack_configs: &verify-2ndline-slack-configs
   - send_resolved: true
     channel: '#verify-2ndline'
     icon_emoji: ':verify-shield:'
     username: alertmanager
 - name: "verify-p1"
   pagerduty_configs:
     - service_key: "${verify_p1_pagerduty_key}"
+  slack_configs: *verify-2ndline-slack-configs
```
issyl0: Hmm, do YAML anchors not work in `.tpl` files even though they're actually YAML, or does Terraform not work like Concourse, ie doesn't expand them in the plan?

`terraform plan` output:

```
name: \"verify-p1\"\n  pagerduty_configs:\n    - service_key: \"[REDACTED]\"\n  slack_configs: *verify-2ndline-slack-configs\n\n"
```

philandstuff: YAML anchors aren't a templating syntax; they're interpreted by alertmanager when it reads the config file.

philandstuff: >  or does Terraform not work like Concourse, ie doesn't expand them in the plan?

so yes i would argue that concourse has the unusual behaviour here (even if it is helpful)

issyl0: Oh, interesting. 👍

PR #283, opened 2019-03-20
Send Verify P2 alerts (integration) to Slack and PagerDuty

issyl0:
https://trello.com/c/0P66rAiU/319-make-integration-alerts-go-to-p2-pagerduty

### DIFF
--- a/terraform/modules/app-ecs-services/templates/alertmanager.tpl
+++ b/terraform/modules/app-ecs-services/templates/alertmanager.tpl
@@ -40,6 +40,10 @@ route:
       match:
         deployment: prod
         severity: p1
+    - receiver: "verify-p2"
+      match:
+        deployment: integration
+        severity: p1
     - match:
         severity: constant
       group_interval: 1m
@@ -101,4 +105,7 @@ receivers:
   pagerduty_configs:
     - service_key: "${verify_p1_pagerduty_key}"
   slack_configs: *verify-2ndline-slack-configs
-
+- name: "verify-p2"
+  pagerduty_configs:
+    - service_key: "${verify_p2_pagerduty_key}"
+  slack_configs: *verify-2ndline-slack-configs
--- a/terraform/modules/app-ecs-services/alertmanager-service.tf
+++ b/terraform/modules/app-ecs-services/alertmanager-service.tf
@@ -158,6 +158,10 @@ data "pass_password" "verify_p1_pagerduty_key" {
   path = "pagerduty/integration-keys/verify-p1"
 }
 
+data "pass_password" "verify_p2_pagerduty_key" {
+  path = "pagerduty/integration-keys/verify-p2"
+}
+
 data "pass_password" "slack_api_url" {
   path = "slack-api-url"
 }


terraform/modules/app-ecs-services/templates/alertmanager.tpl
```diff
@@ -101,4 +105,7 @@ receivers:
   pagerduty_configs:
     - service_key: "${verify_p1_pagerduty_key}"
   slack_configs: *verify-2ndline-slack-configs
-
+- name: "verify-p2"
+  pagerduty_configs:
+    - service_key: "${verify_p1_pagerduty_key}"
```
andy-paine: I don't think this is the right key - it will need to be a different service for p2

issyl0: I'll look into how to make one and where to put it.

terraform/modules/app-ecs-services/templates/alertmanager.tpl
```diff
@@ -40,6 +40,10 @@ route:
       match:
         deployment: prod
         severity: p1
+    - receiver: "verify-p2"
+      match:
+        deployment: integration
+        severity: p2
```
philandstuff: this won't work for us. the logic is:

Alert `HubSamlProxyNonEidasErrorRate` needs to go to Pagerduty P1 when `deployment: prod` but to Pagerduty P2 when `deployment: integration`.  Currently, `HubSamlProxyNonEidasErrorRate` is marked `severity: p1` which goes to Pagerduty in prod and Slack in all other environments.

Now it's probably becoming clear that the `severity: p1` label is misnamed.

Anyway the shortest possible change here is:
```suggestion
        severity: p1
```
which will Do The Right Thing, but it's confusing having a `severity: p1` going to a `verify-p2` receiver, and someone will probably "fix" it at some point, so i think there's a future piece of work to find better words to describe what `severity` is trying to do.

terraform/modules/app-ecs-services/templates/alertmanager.tpl
philandstuff: it lives in re-secrets and we have terraform pass provider data sources to pull secrets out of there

terraform/modules/app-ecs-services/templates/alertmanager.tpl
issyl0: But whatever alert should go to PagerDuty as a P2 if it’s in office hours and in the integration environment, as I understand it.

Currently integration issues show up in Slack a P1 which confuses people (myself included) when we don’t get called about it.

How do we configure in-hours integration alerts as P2 alerts sending to PagerDuty and Slack, and P1 alerts for production, in this case?

philandstuff: @issyl0 that's exactly what my proposed change does:

If it matches `severity:p1` and `deployment:prod`, it goes to pagerduty P1
If it matches `severity:p1` and `deployment:integration`, it goes to pagerduty P2 (which only calls in hours)
If it matches `severity:p1` but has another deployment, it goes to slack.

issyl0:
Things still to do:
[ ] Generate P2 PagerDuty key
[ ] Add it to Secrets

issyl0:
We've not done the "in hours" thing yet, so it could call people out of hours... 🤔

tlwr:
Currently the service is configured to only page using the business hours escalation policy

PR #291, opened 2019-03-27
Continuously Deploy AlertManager

tlwr:


tlwr:
@matthewcullum-gds I've pushed some commits, which notable add the `gpg_private_key` variable and use it. I've got it to the stage that terraform plan works. The commits need some rebasing and we need to get it to apply things.

I also think we can restrict our watching of re-secrets specifically to the `observe` path.

tlwr:
Also we should make it deploy/wait/smoke prod after staging is good.

PR #293, opened 2019-03-28
Add Cloud Security Watch to Staging and Production environments

issyl0:
- Cyber Security have created a tool - Cyber Security Watch - that can
  assume a read-only role to an AWS account and analyse its configuration,
  to identify any security concerns.

### DIFF
--- a/terraform/projects/app-ecs-albs-production/main.tf
+++ b/terraform/projects/app-ecs-albs-production/main.tf
@@ -49,6 +49,12 @@ data "terraform_remote_state" "infra_networking" {
   }
 }
 
+module "csw_role" {
+  source = "git::https://github.com/alphagov/csw-client-role.git?ref=v1.2"
+
+  csw_agent_account_id = "779799343306"
+}
+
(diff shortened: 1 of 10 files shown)


issyl0:
I did this in _all the projects_ because we deploy Terraform for so many different things in their own way. I'm now not convinced this is correct? @tlwr said "once per account"?

tlwr:
If there are 3 AWS accounts this module needs to be instantiated 3 times

What it does is create an IAM role and profile, if it is deployed multiple times per account then deploys will fail

issyl0:
Ah, so not per _project_ but per _account_. Where, then, should I put this? 🤔(We don't generally use the dev account any more, and there shouldn't be anything in it, hence why I've only done staging and prod.)

tlwr:
If you have a project which is deployed once per account? Alternatively: https://github.com/alphagov/gds-user-accounts ?

issyl0:
Superseded by https://github.com/alphagov/gds-user-accounts/pull/11.

PR #296, opened 2019-03-29
Delete all mentions of the dev environment

issyl0:
- We don't do local dev anymore, so we don't need any of the Terraform.


issyl0:
I believe the kitchen tests run in dev, and if we're saying we don't have dev environments any more - should we delete them? 🤔

philandstuff:
@issyl0 yes let's bin them

issyl0:
@philandstuff I removed the kitchen tests. 👩‍🍳

PR #297, opened 2019-03-29
Only report crashes in prod space

gidsg:
We were getting erroneous alerts about staging apps crashing which are not urgent to action.


issyl0:
Deployed. 🚀

PR #299, opened 2019-03-29
bump tf version to latest

philandstuff:


issyl0:
Can we do the Ruby version to 2.6 as well?

philandstuff:
in a separate PR, but yes

issyl0:
We forgot [.travis.yml](https://github.com/alphagov/prometheus-aws-configuration-beta/pull/301).

PR #300, opened 2019-03-29
Better slack notifications

philandstuff:
This is a very scrappy first attempt to improve our slack
notifications.

To do this, it creates a new configuration file, `default.tmpl`, that
contains templates for defining slack notifications.

We absolutely can and should iterate on this.


philandstuff:
Example alert:
image

tools/prometheus-configs/send-an-alert/rules/alerts.yml
```diff
@@ -6,6 +6,14 @@ groups:
...
+      product: verify
     annotations:
       summary: "This is a test alert that always fires."
+      message: |
+        Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed
+        do eiusmod tempor incididunt ut labore et dolore magna
+        aliqua. Ut enim ad minim veniam, quis nostrud exercitation
+        ullamco laboris nisi ut aliquip ex ea commodo consequat. Duis
+        aute irure dolor in reprehenderit in voluptate velit esse
+        cillum dolore eu fugiat nulla pariatur.
```
issyl0: I personally prefer http://www.catipsum.com/.

PR #302, opened 2019-03-29
Bump Ruby to 2.6.1

issyl0:


terraform/modules/prom-ec2/prometheus/.ruby-version
```diff
@@ -1 +1 @@
-2.5.1
+2.6.1
```
philandstuff: Can we delete this if it agreed with the root one?

PR #316, opened 2019-04-30
create a new targets bucket

philandstuff:
Create a new targets bucket.  We're planning to move away from
gds-prometheus-targets{,-staging} (which live in the central
gds-tech-ops account) and have buckets live alongside the prometheus
they supply targets to.

This commit creates a new targets bucket that prometheus can read, and
creates an IAM user that can write to the bucket.

This *doesn't* set up access keys for the new IAM user.  It'll be a
manual step to do that and to set the key up on the cf_app_discovery
instance.

### DIFF
--- a/terraform/modules/prom-ec2/prometheus/targets.tf
+++ b/terraform/modules/prom-ec2/prometheus/targets.tf
@@ -1,0 +1,61 @@
+resource "aws_s3_bucket" "prometheus_targets" {
+  bucket        = "govukobserve-targets-${var.environment}"
+  acl           = "private"
+  force_destroy = true
+
+  versioning {
+    enabled = true
+  }
+}
+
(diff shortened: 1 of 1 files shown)


terraform/modules/prom-ec2/prometheus/targets.tf
```diff
@@ -0,0 +1,61 @@
+resource "aws_s3_bucket" "prometheus_targets" {
+  bucket        = "govukobserve-targets-${var.environment}"
```
andy-paine: Insert comment about how we don't have `Observe` as a concept anymore 😄

philandstuff: yeah i know but naming is hard and we already use this prefix for the config bucket, so consistency or something 😄

PR #320, opened 2019-05-31
Switch the Prometheis to m5.large instances

issyl0:
- We recently discovered that the Prometheis for PaaS (the only ones we
  run via this Terraform) are m4.large. We can save ourselves some cash by
  switching them to m5.large, the next generation.
- We had more variables than we needed for this, which I think were left
  over from when we had ECS EC2 Prometheis, dev stacks, and the "enclave".
- This gets rid of those extra variable settings in
  prom-ec2/paas-<environment> projects and sets the default module
  instance_size variable to m5.large.


andy-paine (approved):

issyl0:
Deployed to staging and discovered we needed ffed8d8 because M5 instances have different device names. It works now we've applied that change too.

sakisv (approved):

andy-paine (approved):

PR #322, opened 2019-05-31
digitalmarketplace _metrics endpoint hack: add "-router" hostname to exceptions list

risicle:
Original hack PR https://github.com/alphagov/prometheus-aws-configuration-beta/pull/268

We now need to be able to do this for our equivalently-named `-router` apps.

:grin: I have clearly not tested this :grin:


andy-paine (approved):
Looks like it should work 🤷‍♂

risicle:
Cool. Any idea when this will go live?

PR #323, opened 2019-06-03
AUTO-413: Add Notify statsd exporter

andy-paine:
- We want to scrape metrics from a statsd exporter that has been stood
  up in AWS for Notify's metrics
- We can't run it on the PaaS due to static IP restrictions and so can't
  use the service broker thus needs an explicit scrape config


issyl0 (approved):

issyl0:
Looks like some YAML.

PR #327, opened 2019-06-21
deploy the service broker from the prometheus pipeline

philandstuff:
because that would be great

### DIFF
--- a/pipeline.yml
+++ b/pipeline.yml
@@ -7,6 +7,11 @@ resources:
       paths:
         - terraform/modules/app-ecs-services
         - terraform/projects/app-ecs-services-*
+  - name: cf-app-discovery-git
+    type: git
+    source:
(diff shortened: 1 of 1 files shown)


pipeline.yml
```diff
@@ -247,4 +268,56 @@ jobs:
...
+              apt-get update
+              # Old versions of the ruby docker set this envvar which
+              # messes things up it was fixed in
+              # docker-library/ruby#255 so we can remove this `unset`
+              # line when we upgrade to ruby 2.6
+              unset BUNDLER_VERSION
+              gem install bundler -v 2.0.1
+              bundle version
+              bundle install --without development
+              rake
```
issyl0:
```suggestion
              bundle exec rake
```

issyl0 (approved):

PR #328, opened 2019-06-21
updates verify gsp cluster egress ip addrs

chrisfarms:
egress IPs changed for verify gsp cluster changed after the
migration to the EKS based deployment a while back, this updates them so
we can wire up alertmanager again.


issyl0 (dismissed):
I trust that these are correct - I don't think I have any way to check?

andy-paine (approved):
Assume these are correct too but ```Error running fmt: In ./terraform/projects/infra-security-groups-production/main.tf: At 91:5: error parsing list, expected comma or list end, got: STRING
The command "find . -name '*.tf' | xargs tools/terraform-format.sh" exited with 123.``` from Travis

terraform/projects/infra-security-groups-production/main.tf
```diff
@@ -86,10 +86,10 @@ module "infra-security-groups" {
     "35.176.252.164/32",
 
     # verify gsp
-    "35.176.76.151/32",
+    "18.130.62.225"
```
issyl0:
```suggestion
    "18.130.62.225",
```

terraform/projects/infra-security-groups-production/main.tf
```diff
@@ -86,10 +86,10 @@ module "infra-security-groups" {
...
 
-    "18.130.183.50/32",
-    "18.130.4.61/32",
+    "3.8.38.79"
```
issyl0:
```suggestion
    "3.8.38.79",
```

terraform/projects/infra-security-groups-production/main.tf
```diff
@@ -86,10 +86,10 @@ module "infra-security-groups" {
...
+    "52.56.126.222"
```
issyl0:
```suggestion
    "52.56.126.222",
```

philandstuff:
I suspect the IP addresses *must* end with `/32` or it will break 😞

issyl0 (approved):

PR #329, opened 2019-06-26
Remove DM relabelling hacks

philandstuff:
as of alphagov/cf_app_discovery#43, the service broker now outputs
correct metrics paths so we don't need to put hacks in here to set the
correct metrics path.

### DIFF
--- a/terraform/modules/prom-ec2/paas-config/prometheus.conf.tpl
+++ b/terraform/modules/prom-ec2/paas-config/prometheus.conf.tpl
@@ -34,30 +34,6 @@ scrape_configs:
     file_sd_configs:
       - files: ['/etc/prometheus/targets/*.json']
         refresh_interval: 30s
-    # Digital marketplace's API and frontend apps listen on a
-    # different metrics path, so when we detect one we modify the
-    # __metrics_path__ label to set the path we scrape
-    relabel_configs:
-      - source_labels: ['__metrics_path__', 'org', 'job']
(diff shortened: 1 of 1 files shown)


terraform/modules/prom-ec2/paas-config/prometheus.conf.tpl
```diff
@@ -34,30 +34,6 @@ scrape_configs:
...
-        target_label: '__metrics_path__'
-        replacement: '$1/_metrics'
-      - source_labels: ['org', 'job']
-        regex: 'digitalmarketplace;router(-release)?'
-        target_label: '__metrics_path__'
-        replacement: '/_metrics'
-      - source_labels: ['org', 'space', 'job']
-        regex: 'digitalmarketplace;(.*);router(-release)?'
-        target_label: '__address__'
-        replacement: 'dm-router-$1.cloudapps.digital'
```
Krenair: What is this last one?

andy-paine: Extra hacks for the router to try and deal with the multiple domains thing I think - IIRC it has a normal `/metrics` endpoint but the address was getting confused due to sometimes having a -release and also having a custom route attached to it. The PR to discover the routes with a `/metrics` should solve this issue too

andy-paine (approved):
I feel like you've been wanting to remove this since the moment you added it 🚀

philandstuff:
Deployed and working 🚀

PR #332, opened 2019-07-03
Alerts tidying

philandstuff:
See commits for context.  Basically making the autom8 alerts channel nicer.


issyl0 (approved):
Seems sensible, and consistency and tidiness are both good things. I'd like to see it in practice, but we can't do that until we deploy it so :+1:.

PR #335, opened 2019-07-05
[slack] warnings should be warning-colored, not danger-colored

philandstuff:
This tweaks the color of a slack alert to match the emoji.


issyl0 (approved):
I never knew alerts could have colours other than red or green.

Krenair (approved):

PR #336, opened 2019-07-09
Add London targets configuration

andy-paine:
- We want to be able to scrape PaaS targets in London PaaS orgs
- Create a new bucket for the file_sd_configs and add the cron
  job/scrape config to pick up new targets
- Might need the bucket applying first and folders called active
  and inactive creating in it

### DIFF
--- a/terraform/modules/prom-ec2/paas-config/prometheus.conf.tpl
+++ b/terraform/modules/prom-ec2/paas-config/prometheus.conf.tpl
@@ -34,6 +34,12 @@ scrape_configs:
     file_sd_configs:
       - files: ['/etc/prometheus/targets/*.json']
         refresh_interval: 30s
+  - job_name: paas-london-targets
+    scheme: http
+    proxy_url: 'http://localhost:8080'
+    file_sd_configs:
+      - files: ['/etc/prometheus/targets/london/*.json']
+        refresh_interval: 30s
   - job_name: alertmanager
     dns_sd_configs:
       - names:
--- a/terraform/modules/prom-ec2/paas-config/prometheus.conf.tpl
+++ b/terraform/modules/prom-ec2/paas-config/prometheus.conf.tpl
@@ -40,6 +40,9 @@ scrape_configs:
     file_sd_configs:
       - files: ['/etc/prometheus/targets/london/*.json']
         refresh_interval: 30s
+    relabel_configs:
+      - target_label: region
+        replacement: london
   - job_name: alertmanager
     dns_sd_configs:
       - names:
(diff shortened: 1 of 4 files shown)


terraform/modules/prom-ec2/paas-config/prometheus.conf.tpl
```diff
@@ -34,6 +34,12 @@ scrape_configs:
     file_sd_configs:
       - files: ['/etc/prometheus/targets/*.json']
         refresh_interval: 30s
+  - job_name: paas-london-targets
+    scheme: http
+    proxy_url: 'http://localhost:8080'
+    file_sd_configs:
+      - files: ['/etc/prometheus/targets/london/*.json']
+        refresh_interval: 30s
```
philandstuff: can we have some `relabel_configs` here to add `region: london` to everything?

terraform/modules/prom-ec2/prometheus/main.tf
```diff
@@ -66,14 +66,15 @@ data "template_file" "user_data_script" {
...
-    region              = "${var.region}"
-    targets_bucket      = "${aws_s3_bucket.prometheus_targets.id}"
-    alerts_bucket       = "${aws_s3_bucket.prometheus_config.id}"
-    prom_external_url   = "https://${var.prometheus_public_fqdns[count.index]}"
-    logstash_host       = "${var.logstash_host}"
-    prometheus_htpasswd = "${var.prometheus_htpasswd}"
-    allowed_cidrs       = "${join("\n        ",formatlist("allow %s;", var.allowed_cidrs))}"
+    config_bucket         = "${aws_s3_bucket.prometheus_config.id}"
+    region                = "${var.region}"
+    targets_bucket        = "${aws_s3_bucket.prometheus_targets.id}"
```
philandstuff: I'd like to rename this to `ireland_targets_bucket` at some point but not needed now

philandstuff (approved):
This is a great start and probably good to merge now; comments below could be addressed now or in future PRs

terraform/modules/prom-ec2/paas-config/prometheus.conf.tpl
andy-paine: Couldn't work out how to do this - realised you can add labels with relabel too. See next commit.

issyl0:
This will require the instances to respin, but I can do that tomorrow morning (or later tonight) if you like?

andy-paine:
`Tonight` is never a good time to be spinning stuff :grimacing: Let's go with tomorrow if you would be so kind :+1: 

I think we might need to make the folders in the bucket first as I think the cron just checks for if the bucket exists, not if the folders exist in the bucket. Easy fix forward though if it doesn't work.

PR #340, opened 2019-07-12
Fix volume mounting

Krenair:
- When we re-deployed the Prometheis, we found that their volumes didn't attach unless we manually attached them. The cloud-init-output.log was complaining about `nvme0 is not a block device`. This appears to be very similar to the fix Autom8 had to do for Verify - alphagov/verify-infrastructure#235 - except we've also found that mounting /dev/nvme0 on first boot no longer works, we need /dev/nvme1n1.
- We suspect that Amazon changed something recently, as this was previously working fine even after we switched to m5 instances from m4s.

Co-Authored-By: Isabell Long <isabell.long@digital.cabinet-office.gov.uk>


issyl0 (approved):
This is working.

PR #344, opened 2019-09-10
Bump Prometheus data volumes

samcrang:
We're close to running out of disk on 2/3 `prom`s. We have run out
entirely of disk on `prom-1`. The disk usage appears to be accelerating
even though the rate of metric ingestion seems largely unchanged. We're
going to bump the disk size for the time being.

### DIFF
--- a/terraform/modules/prom-ec2/prometheus/main.tf
+++ b/terraform/modules/prom-ec2/prometheus/main.tf
@@ -53,7 +53,7 @@ resource "aws_ebs_volume" "prometheus-disk" {
   count = "${length(keys(var.availability_zones))}"
 
   availability_zone = "${element(keys(var.availability_zones), count.index)}"
-  size              = "21"
+  size              = "50"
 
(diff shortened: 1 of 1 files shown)


terraform/modules/prom-ec2/prometheus/main.tf
```diff
@@ -53,7 +53,7 @@ resource "aws_ebs_volume" "prometheus-disk" {
   count = "${length(keys(var.availability_zones))}"
 
   availability_zone = "${element(keys(var.availability_zones), count.index)}"
-  size              = "21"
+  size              = "30"
```
issyl0: Is this enough?

Krenair (approved):
I was going to go for 25 and see how we got on, but 30 works too

samcrang:
5️⃣0️⃣

issyl0 (approved):

Krenair (approved):

PR #347, opened 2019-09-17
send DCS and eIDAS alerts to separate channels

philandstuff:
Also, refactor the alertmanager routing tree

Also also, ensure sandbox and verify GSP alerts both get the GSP-style formatting.

### DIFF
--- a/terraform/modules/app-ecs-services/templates/alertmanager.tpl
+++ b/terraform/modules/app-ecs-services/templates/alertmanager.tpl
@@ -43,6 +43,31 @@ route:
     match:
       layer: "infra"
       severity: "ticket"
+  # GSP clusters
+  - match_re:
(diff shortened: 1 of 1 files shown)


terraform/modules/app-ecs-services/templates/alertmanager.tpl
```diff
@@ -153,7 +161,8 @@ receivers:
       url: '{{ .CommonAnnotations.runbook_url }}'
 - name: "autom8-gsp-alerts-slack"
   slack_configs:
-  - send_resolved: true
+  - &GSP-slack-config
```
issyl0: Can we lowercase this for consistency with the rest of the _thing_ names? (eidas, autom8, dcs)

adityapahuja (approved):
⭐️

PR #348, opened 2019-09-19
Restrict Proxy Node alerts to the prod and integration namespaces

idrop:
We're getting slack alerts when our gsp deployments are not healthy.

To ensure we care about these alerts, we should only get them for the integration and prod namespaces.

This PR will limit alerts to these namespaces.


Wynndow (approved):

philandstuff (changes requested):
This will put the rest of them back into the default receiver (which is #re-autom8-alerts).  Could you add a rule that puts your uninteresting alerts into the `dev-null` receiver instead?

tommrr (approved):

instantiator (approved):

philandstuff (approved):

PR #353, opened 2019-11-14
Add Verify GSP canary target.

blairboy362:
We've been seeing intermittent connection failures from multi-tenant
concourse to the various canary deployments. We'd like to add a prometheus
target so we can gather more data about the state of the canary in the
Verify cluster.


chrisfarms (approved):

philandstuff:
post hoc ✅

PR #363, opened 2020-01-06
BAU: Revert dcs prod only.

Todai88:
As we are adding a new alert that will make sure that developers are alerted if any of our pipelines are red, we can no longer restrict alerts to the prod-namespace.

### DIFF
--- a/terraform/modules/app-ecs-services/templates/alertmanager.tpl
+++ b/terraform/modules/app-ecs-services/templates/alertmanager.tpl
@@ -61,8 +61,8 @@ route:
     - match:
         severity: constant
       receiver: "dev-null"
-    - match:
-        namespace: verify-doc-checking-prod
+    - match_re:
+        namespace: verify-doc-checking-*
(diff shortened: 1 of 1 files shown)


terraform/modules/app-ecs-services/templates/alertmanager.tpl
```diff
@@ -62,7 +62,7 @@ route:
         severity: constant
       receiver: "dev-null"
     - match:
```
chrisfarms: match will need to be match_re I think to make use of regex

chrisfarms (changes requested):

terraform/modules/app-ecs-services/templates/alertmanager.tpl
Todai88: 👍 ?

chrisfarms (approved):

chrisfarms:
bring the noise!

PR #365, opened 2020-01-06
BAU: add routes for DCS alerts

Todai88:
Added two routes for potential DCS alerts; they need to either:
* Be sent from the `prod` namespace or,
* Have the label and value `type: pipeline-alert`.

If none of them match, the alert is discarded.

This way we can get pipeline alerts from other namespaces than the `prod` namespace.

### DIFF
--- a/terraform/modules/app-ecs-services/templates/alertmanager.tpl
+++ b/terraform/modules/app-ecs-services/templates/alertmanager.tpl
@@ -62,8 +62,15 @@ route:
         severity: constant
       receiver: "dev-null"
     - match_re:
-        namespace: verify-doc-checking-.*
-      receiver: dcs-slack
+          namespace: verify-doc-checking-.*
+        receiver: dev-null
+        routes:
+        - receiver: dcs-slack
+          match: 
+            type: pipeline-alert
+        - receiver: dcs-slack
+          match:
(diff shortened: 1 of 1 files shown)


terraform/modules/app-ecs-services/templates/alertmanager.tpl
```diff
@@ -63,7 +63,13 @@ route:
       receiver: "dev-null"
     - match_re:
```
chrisfarms: I think (but im not sure) that you might need to add a dev-null receiver for the ones that don't match... like:

```
      - match_re:
          namespace: verify-doc-checking-.*
        receiver: dev-null
        routes:
        - receiver: dcs-slack
          match: 
            type: pipeline-alert
        - receiver: dcs-slack
          match:
            namespace: verify-doc-checking-prod
```

which I think will:

* route to dcs-slack if type:pipeline-alert AND namespace:verify-doc-checking-*
* route to dcs-slack if namespace:verify-doc-checking-prod AND namespace:verify-doc-checking-*
* else route to dev-null 

Maybe it bubbles up to the next receiver up, but I'm not 100% sure

chrisfarms (approved):

chrisfarms (approved):

PR #366, opened 2020-01-07
BAU: Re-add prod only route

Todai88:
Also:
	- Remove old routes (non-prod)

## Why?

It seems too difficult to get the DCS pipeline alerts to be caught by the `AlertManager` this way. Will instead implement using the `slack-notifaction` task directly in the pipeline.

### DIFF
--- a/terraform/modules/app-ecs-services/templates/alertmanager.tpl
+++ b/terraform/modules/app-ecs-services/templates/alertmanager.tpl
@@ -61,16 +61,9 @@ route:
     - match:
         severity: constant
       receiver: "dev-null"
-    - match_re:
-          namespace: verify-doc-checking-.*
-        receiver: dev-null
-        routes:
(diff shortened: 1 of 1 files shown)


chrisfarms (approved):

idrop (commented):
Should you send the uninteresting matches to the `dev-null` receiver? I did this previously and got this feedback: https://github.com/alphagov/prometheus-aws-configuration-beta/pull/348

Todai88:
> Should you send the uninteresting matches to the `dev-null` receiver? I did this previously and got this feedback: #348

We _probably_ should - but I haven't heard anything from RE about alerts being popped into their channel (and I haven't seen any in it either). We haven't historically been sending anything to `dev-null`, that's something I added yesterday because we weren't restricting alerts to the `prod` namespace.
 
@chrisfarms any clues? Happy to add it if necessary!

PR #367, opened 2020-01-24
upgrade to terraform 0.12

philandstuff:
Pretty dumb upgrade, this is mostly a no-op except for the fact there's a new ECS AMI so applying prom-ec2 will rebuild prometheus.


.terraform-version
```diff
@@ -1 +1 @@
-0.11.13
+0.12.19
```
issyl0: 0.12.20 is out now. 🙈

philandstuff: don't care, wasn't when i started this 😄

szd55gds (changes requested):
Hey @philandstuff , what a monster! My only concern is that in terraform/projects/infra-security-groups-production/main.tf, you remove comments which ID IP addresses to various GDS projects/services, I think this is a retrograde step because I can see folk in the near future asking "what on earth is that IP". Suggest we leave those comments in!

philandstuff:
Good spot @szd55gds, I've pushed a commit to reinstate them

szd55gds (approved):
Looks magnificent to me :-)

PR #371, opened 2020-01-27
Fix prometheus EC2 service discovery

philandstuff:
I removed the Job tag and now prometheus can't scrape itself, which is
causing alerts..


philandstuff:
See #370 for where the Job tag went

szd55gds (approved):
Looks ok to me

PR #372, opened 2020-01-28
add NLB-based alertmanager 

philandstuff:
This adds a new alertmanager deployment, to supersede the existing
one, based on an NLB instead of an ALB.

Why?  Because it's simpler.  The main benefit is that NLBs support
same-zone routing, so that each public IP address will route traffic
only to things in the same AZ.  This means that each public IP will
reliably route to the same alertmanager instance.

This means we don't need clever listener rules to inspect the `Host`
header to route traffic.

A separate but related change: we can also create A records with the
same name as each other, and let Prometheus discover Alertmanager
using dns_sd_configs.  Then Prometheus will send each alert to each of
the NLB IP addresses, and therefore to each alertmanager.

That is, instead of `alerts-1.gds-reliability.engineering` and similar
for alerts-2 and alerts-3, we can have plain
`alerts.gds-reliability.engineering` with 3 separate A records on it.

(It's worth spelling out why we want an NLB at all rather than just
running alertmanager on a public subnet: an NLB will terminate TLS for
us, using ACM certificates.  That's basically the only reason we want
the NLB here.)

One downside is that NLBs don't support having a 80->443 redirect.  I
think this is a problem we can live with.

Once this change is complete, it would also mean that alertmanager is
no longer dependent on the app-ecs-albs module.  We could rename
modules to make them more meaningful:

 - app-ecs-services -> alertmanager
 - app-ecs-albs -> prometheus-alb (it would only have a single ALB at
   this point)
 - infra-security-groups -> prometheus-security-groups


tlwr (approved):
this is clever (in a good way)

samcrang (approved):

PR #384, opened 2020-01-31
[alertmanager] SANs on the cert for the new alertmanager DNS records

philandstuff:
oops, the smoke tests won't pass without this


szd55gds (approved):
looks ok

PR #385, opened 2020-04-07
Detect device identifier of prometheus EBS volume for mounting.

blairboy362:
EBS volumes are mapped to device identifiers in the kernel in a way that
is unpredictable (it gives numbers and letters based on the order in which
it becomes aware of them and so it cannot be relied upon). We are seeing
an issue whereby the ordering of the devices sometimes reverses and so an
attempt is made to mount the wrong device.

Also, the check for whether the disk was formatted was failing sometimes.
The `file -s` command wasn't returning the expected string and so a disk
that needed formatting wasn't being formatted. The mount command that
follows then fails because there's no filesystem to mount.

This change attempts to discover the identifier by looking it up by disk
size. It's a hack, but, for the moment, prometheus is only using a single
disk of that size. The failing `file -s` check has been replaced by a
check interrogating `blkid` which seems to be more reliable in testing.

### DIFF
--- a/terraform/modules/prom-ec2/prometheus/cloud.conf
+++ b/terraform/modules/prom-ec2/prometheus/cloud.conf
@@ -33,19 +33,35 @@ write_files:
         # if alerts bucket exists then sync it, otherwise this cron runs but has no effect
         * * * * * root [ "${alerts_bucket}" != "" ] && aws s3 sync s3://${alerts_bucket}/prometheus/alerts/ /etc/prometheus/alerts --region=${region} --delete
   - content: |
-       #!/bin/bash
-       if file -s /dev/nvme1n1 | grep -q "/dev/nvme1n1: data"; then
-         mkfs -t 'ext4' -L 'prometheus_disk' '/dev/nvme1n1'
-       else
-         echo "disk already formatted"
-       fi
-    path: /root/format_disk.sh
-    permissions: 0755
-  - content: |
-       #!/bin/bash
-       UUID=$(blkid /dev/nvme1n1 -s UUID -o value)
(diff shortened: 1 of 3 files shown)


blairboy362:
This has been successfully applied to staging.

terraform/modules/prom-ec2/prometheus/cloud.conf
```diff
@@ -33,19 +33,35 @@ write_files:
...
-       UUID=$(blkid /dev/nvme1n1 -s UUID -o value)
-       echo "UUID=$UUID /mnt ext4 defaults,nofail 0 2" >> /etc/fstab
-    path: /root/generate_fstab.sh
+        echo 'Configuring prometheus EBS'
+        vol=""
+        while [ -z "$vol" ]; do
+          # adapted from
+          # https://medium.com/@moonape1226/mount-aws-ebs-on-ec2-automatically-with-cloud-init-e5e837e5438a
+          # [Last accessed on 2020-04-02]
+          vol=$(lsblk | grep -e disk | awk '{sub("G","",$4)} {if ($4+0 == ${data_volume_size}) print $1}')
```
Krenair, line 42: what's the purpose of the `+0`? cast to integer or something?

Krenair (commented):
looks good, one tiny question

terraform/modules/prom-ec2/prometheus/cloud.conf
blairboy362: Yes. It was used in the example I copied from and couldn't find a decent description in the awk documentation. As usual, though. stackoverflow helped:
https://stackoverflow.com/questions/5808971/casting-to-int-in-awk

Krenair (approved):

PR #388, opened 2020-05-13
Continuous deployment pipeline for prom-ec2 project

chrisfarms:
## What

Continuously deploy prom-ec2 project from master

* Splits out terraform task so it can be shared
* Deploys the prom-ec2 staging project ... performs a simple curl check to see if it's up
* Repeats for production
* Builds a task image to avoid having to do lots of apt-get still in every single task run
* Adds a set_pipeline step so pipeline is also continuously deployed.

## Why

Because deploying by hand is a faff and since alertmanager config is already continuously deployed it causes confusion.

## Pipeline

You can see the pipeline here: https://cd.gds-reliability.engineering/teams/autom8/pipelines/prometheus

It is already set-piplined ... but the configure job is paused as it will error until this is merged.

## Note

This deploys all three in one big go... I seriously considered making it target them one-by-one ... but it just isn't very practical because: we do not know the alert config terraform "target" (they are dynamic / new ones get added) ... so we cannot "target" just that part

One improvement could be to try "just one" without any of the config changes, "just to check" before moving on to all three .... but I think that is just going to confuse things.

### DIFF
--- a/ci/images/task/Dockerfile
+++ b/ci/images/task/Dockerfile
@@ -1,0 +1,40 @@
+FROM ubuntu:18.04
+
+ENV TF_VERSION 0.12.19
+
+LABEL ubuntu="18.04"
+LABEL terraform="$TF_VERSION"
+
+ENV TZ=Europe/London
+RUN ln -snf /usr/share/zoneinfo/$TZ /etc/localtime && echo $TZ > /etc/timezone
+
+RUN apt-get update  --yes && \
+	apt-get install --yes --no-install-recommends \
+	ca-certificates \
+	awscli \
+	jq \
+	curl \
+	dnsutils \
+	unzip \
+	gpg \
+	gpg-agent \
+	golang \
+	git
+
+WORKDIR /tmp
+
+RUN curl https://releases.hashicorp.com/terraform/${TF_VERSION}/terraform_${TF_VERSION}_linux_amd64.zip > terraform.zip && \
+	echo 'a549486112f5350075fb540cfd873deb970a9baf8a028a86ee7b4472fc91e167  terraform.zip'  > terraform.sha && \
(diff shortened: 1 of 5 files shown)


ci/tasks/http-ping.yml
```diff
@@ -0,0 +1,19 @@
...
+    tag: fe3e384e81ccb50842509d7237e3828b293de694
+params:
+  URL:
+run:
+  path: sh
+  args:
+    - -euxc
+    - |
+      DOMAIN=$(echo "${URL}" | awk -F/ '{print $3}')
+      getent ahosts ${DOMAIN} | cut -d ' ' -f1 | sort | uniq | xargs -n1 -I {} curl \
```
blairboy362, line 15: I have no idea what this line does. I'm sure you've tested it and it works a treat! :)

blairboy362 (approved):

ci/images/task/Dockerfile
```diff
@@ -0,0 +1,40 @@
...
+	rm terraform.zip && rm terraform.sha
+
+RUN go get -v github.com/camptocamp/terraform-provider-pass && \
+	mkdir -p ~/.terraform.d/plugins/linux_amd64 && \
+	mv ~/go/bin/terraform-provider-pass ~/.terraform.d/plugins/linux_amd64/
+
+# prom-ec2 terraform expects a pub ssh key even if it doesn't use it
+RUN mkdir -p $HOME/.ssh/ && touch $HOME/.ssh/id_rsa.pub
+
+COPY assume-role /usr/bin/assume-role
```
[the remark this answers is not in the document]
chrisfarms: is that required? ... git stores attributes fine doesn't it?

ci/tasks/http-ping.yml
chrisfarms: yeah name service voodoo ... it's like a poor man's `dig`

PR #394, opened 2020-06-16
Fix observe alert

philandstuff:
We have an alert for missing targets from the service broker, to
detect if there's a problem with the plumbing from the service broker,
via the S3 bucket, to the prometheus targets.

This used to be based on detecting the absence of the metric
`prometheus_sd_file_timestamp`.  This has broken with the upgrade to
focal, bringing prometheus 2.15 with it, because the metric has been
renamed to `prometheus_sd_file_mtime_seconds`.

However, rather than use this metric, we can now use a better way of
solving this problem, which is the `prometheus_sd_discovered_targets`
metric, which actually tells us how many targets we have detected
directly.  This is better than using `absent()` because the metric is
more explicit about what we mean.


chrisfarms (approved):
:+1:

PR #397, opened 2020-07-01
increase prometheus volume size to 250G

chrisfarms:
because we ran out of space.

adds `resize2fs` call to size volume sizing on boot

### DIFF
--- a/terraform/modules/prom-ec2/prometheus/variables.tf
+++ b/terraform/modules/prom-ec2/prometheus/variables.tf
@@ -7,7 +7,7 @@ variable "device_mount_path" {
 
 variable "data_volume_size" {
   description = "The size of the volume that will contain the prometheus data"
-  default     = 50
+  default     = 250
 }
 
 variable "availability_zones" {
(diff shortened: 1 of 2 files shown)


blairboy362 (approved):

terraform/modules/prom-ec2/prometheus/variables.tf
```diff
@@ -7,7 +7,7 @@ variable "device_mount_path" {
 
 variable "data_volume_size" {
   description = "The size of the volume that will contain the prometheus data"
-  default     = 50
+  default     = 250
```
philandstuff, line 10: can we only change this in prod, not in staging too?

chrisfarms: I want to test the resize logic works on way through staging ... but can lower it in PR after

philandstuff: just set the variable in terraform/projects/prom-ec2/paas-production/main.tf

philandstuff (approved):

PR #403, opened 2020-07-21
s9SNoTta: config to route dcs p2 alerts to pagerduty

dbes-gds:
Add new receiver and key for dcs p2 alerts.

### DIFF
--- a/terraform/modules/alertmanager/alertmanager-service.tf
+++ b/terraform/modules/alertmanager/alertmanager-service.tf
@@ -134,6 +134,10 @@ data "pass_password" "verify_p2_pagerduty_key" {
   path = "pagerduty/integration-keys/verify-p2"
 }
 
+data "pass_password" "dcs_p2_pagerduty_key" {
+  path = "pagerduty/integration-keys/dcs-p2"
+}
+
 data "pass_password" "slack_api_url" {
   path = "slack-api-url"
 }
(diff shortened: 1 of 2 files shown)


terraform/modules/alertmanager/alertmanager-service.tf
```diff
@@ -175,6 +175,7 @@ data "template_file" "alertmanager_config_file" {
     govuk_pagerduty_key     = data.pass_password.govuk_pagerduty_key.password
     verify_p1_pagerduty_key = data.pass_password.verify_p1_pagerduty_key.password
     verify_p2_pagerduty_key = data.pass_password.verify_p2_pagerduty_key.password
+    dcs_p2_pagerduty_key    = data.pass_password.dcs_p2_pagerduty_key.password
```
chrisfarms, line 178: you'll need to add the `data` resource for this above .... see ~ line 130 to copy what one of the others looks like

chrisfarms (changes requested):

terraform/modules/alertmanager/alertmanager-service.tf
dbes-gds: done

terraform/modules/alertmanager/templates/alertmanager.tpl
```diff
@@ -67,6 +67,10 @@ route:
     - match_re:
         namespace: verify-doc-checking-prod|verify-dcs-integration
       receiver: dcs-slack
+    - match:
+        namespace: verify-doc-checking-prod
+        severity: p2
+      receiver: "dcs-p2"
```
chrisfarms, line 73: Currently this will NOT be caught as the matcher above it will catch all the things that match "namespace = verify-doc-checking-prod" and send them to slack.

I think what you want is to add some sub routing _below_ that matcher, that further filters on the prod namespace and the p2 severity label.

Like:

```suggestion
      routes:
      - match:
          namespace: verify-doc-checking-prod
          severity: p2
        receiver: "dcs-p2"
```

chrisfarms (approved):

PR #404, opened 2020-07-28
Scrape PaaS's alpha Elasticsearch metrics endpoint

46bit:
## What

PaaS have deployed a new metrics endpoint which collects all metrics for Digital Marketplace's Elasticsearch instances. This PR adds the new scrape target to the Prometheus config and uses a `pass` secret as the password for the scrape endpoint.

The scrape config for this endpoint is very large because there are 700+ metrics exposed and we only want to import metrics which we expect to be useful. Future versions of the Elasticsearch exporter will try to improve on that.

## How to review

I can't easily deploy this, so I suppose we'll can deploy it to staging and see if it works.

~There's a new secret that will need to go in your `pass` store. Add me to your `pass` store or message me privately to get that done.~ Done.


terraform/projects/prom-ec2/paas-production/main.tf
```diff
@@ -64,6 +64,10 @@ data "pass_password" "prometheus_htpasswd" {
   path = "prometheus-basic-auth-htpasswd"
 }
 
+data "pass_password" "dm_elasticsearch_metrics_password" {
+  path = "dm-elasticsearch-metrics-password"
```
46bit, line 68: We'll need to set this path correctly once the new secret is in your `pass` store.

46bit:
Ready for review.

terraform/projects/prom-ec2/paas-production/main.tf
46bit: Done in https://github.com/alphagov/re-secrets/commit/75c91ca4e0051f70d9625f3da71a9af2484d5a5c and this won't need changing.

terraform/modules/prom-ec2/paas-config/prometheus.conf.tpl
```diff
@@ -69,3 +69,163 @@ scrape_configs:
...
       - targets: ['canary.verify-main.london.verify.govsvc.uk']
+  - job_name: paas_elasticsearch_for_dm
+    scheme: https
+    basic_auth:
+      username: digitalmarketplace
+      password: ${dm_elasticsearch_metrics_password}
+    metrics_path: '/federate'
+    params:
+      "match[]":
+      - "{job='aiven'}"
```
philandstuff, line 80: Rather than grabbing all aiven metrics and filtering with `metric_relabel_configs`, could we ask the federate endpoint to only provide the metrics we need?  For example:
```suggestion
      - "disk_free{job='aiven'}"
      - "disk_used_percent{job='aiven'}"
      - "{__name__=~"system_load.*",job='aiven'}"
      - "{__name__=~"elasticsearch_clusterstats.*",job='aiven'}"
```

This will make less work for prometheus since it doesn't have to filter out a bunch of metrics on every scrape.

My only concern is that under the hood, this will construct a giant query string for the `/federate` request, and it may end up with a query that's too large.  I think my answer to that would be multiple scrape_configs to scrape different sets of metrics.

terraform/modules/prom-ec2/paas-config/prometheus.conf.tpl
```diff
@@ -69,3 +69,163 @@ scrape_configs:
     scheme: https
     static_configs:
       - targets: ['canary.verify-main.london.verify.govsvc.uk']
+  - job_name: paas_elasticsearch_for_dm
```
philandstuff, line 72: I think I'd rather pass the whole scrape_config into the module as one big string, with a default of the empty string.  We'd then template and construct the scrape_config in `terraform/projects/prom-ec2/paas-production`.

Why?

 - so that our staging observe prometheus doesn't try to scrape this stuff
 - to keep explicit references to DM out of the supposedly-generic prom-ec2 module

46bit: This is a good idea. I'll restructure the PR to do it. 👍

terraform/modules/prom-ec2/paas-config/prometheus.conf.tpl
46bit: This is a good idea but I'd like to avoid changing it for now. This exporter is probably very temporary and in future we won't need to filter out metrics in this way.

46bit:
@philandstuff Added an extra commit doing what you suggested. There's now a lot more moving parts and it'll need trying out by someone with access to the staging environment.

terraform/projects/prom-ec2/paas-production/main.tf
```diff
@@ -64,6 +64,18 @@ data "pass_password" "prometheus_htpasswd" {
...
+}
+
+data "template_file" "extra_prometheus_scrape_configs_yaml" {
+  template = file("${path.module}/extra-prometheus-scrape-configs.yml.tpl")
+
+  vars = {
+    dm_elasticsearch_metrics_password = data.pass_password.dm_elasticsearch_metrics_password.password
+  }
+}
+
```
philandstuff, line 78:
```suggestion
```
`data.template_file` is deprecated, we should migrate to the tf 0.12 `templatefile` function instead.  (Yes I know data.template_file is used elsewhere here, we haven't got around to fixing it)

terraform/projects/prom-ec2/paas-production/main.tf
```diff
@@ -102,6 +114,8 @@ module "paas-config" {
   prom_private_ips  = module.prometheus.private_ip_addresses
   private_zone_id   = data.terraform_remote_state.infra_networking.outputs.private_zone_id
   private_subdomain = data.terraform_remote_state.infra_networking.outputs.private_subdomain
+
+  extra_scrape_configs = yamldecode(template_file.extra_prometheus_scrape_configs_yaml.rendered)
```
philandstuff, line 118:
```suggestion
  extra_scrape_configs = yamldecode(templatefile("${path.module}/extra-prometheus-scrape-configs.yml.tpl", {
    dm_elasticsearch_metrics_password = data.pass_password.dm_elasticsearch_metrics_password.password
  }))
```

philandstuff (approved):

PR #408, opened 2020-08-03
Escape Prometheus variable substitution

46bit:
Still trying to fix issues not anticipated in #404. 🤦 

This template uses Terraform templatefile variable substitution to create a Prometheus config. Prometheus will then do its own variable interpolation when interpreting the `metric_relabel_configs`. This commit fixes the Prometheus interpolation line so that Terraform won't perform (failed) substitution [1].

[1] https://github.com/hashicorp/terraform/issues/17899#issuecomment-382956001


philandstuff (approved):
The plan makes this change:

```diff
-     "replacement": "paas_es_1"
+     "replacement": "paas_es_${1}"
```

LGTM

PR #409, opened 2020-08-06
Fix list vs set usage

blairboy362:
Sets cannot be indexed directly. Use the recommended for_each construct
instead.

### DIFF
--- a/terraform/modules/alertmanager/nlb.tf
+++ b/terraform/modules/alertmanager/nlb.tf
@@ -30,18 +30,28 @@ resource "aws_acm_certificate" "alertmanager_cert" {
 }
 
 resource "aws_route53_record" "alertmanager_cert_validation" {
-  count      = 1 + length(data.aws_availability_zones.available.names)
-  name       = aws_acm_certificate.alertmanager_cert.domain_validation_options[count.index]["resource_record_name"]
-  type       = aws_acm_certificate.alertmanager_cert.domain_validation_options[count.index]["resource_record_type"]
-  zone_id    = local.zone_id
-  records    = [aws_acm_certificate.alertmanager_cert.domain_validation_options[count.index]["resource_record_value"]]
-  ttl        = 60
+  for_each = {
+    for dvo in aws_acm_certificate.alertmanager_cert.domain_validation_options : dvo.domain_name => {
+      name   = dvo.resource_record_name
+      record = dvo.resource_record_value
+      type   = dvo.resource_record_type
+    }
+  }
+
+  name    = each.value.name
+  records = [each.value.record]
+  type    = each.value.type
+  zone_id = local.zone_id
+  ttl     = 60
+
+  allow_overwrite = true
+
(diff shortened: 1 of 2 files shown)


terraform/modules/alertmanager/nlb.tf
```diff
@@ -30,18 +30,26 @@ resource "aws_acm_certificate" "alertmanager_cert" {
 }
 
 resource "aws_route53_record" "alertmanager_cert_validation" {
-  count      = 1 + length(data.aws_availability_zones.available.names)
-  name       = aws_acm_certificate.alertmanager_cert.domain_validation_options[count.index]["resource_record_name"]
-  type       = aws_acm_certificate.alertmanager_cert.domain_validation_options[count.index]["resource_record_type"]
-  zone_id    = local.zone_id
-  records    = [aws_acm_certificate.alertmanager_cert.domain_validation_options[count.index]["resource_record_value"]]
-  ttl        = 60
+  for_each = {
```
blairboy362, line 33: I'm not entirely convinced this will match the `1 +` logic so I'd appreciate someone familiar with this to check this carefully.

terraform/modules/app-ecs-albs/main.tf
```diff
@@ -106,20 +106,26 @@ resource "aws_acm_certificate" "prometheus_cert" {
...
 resource "aws_route53_record" "prometheus_cert_validation" {
-  # Count matches the domain_name plus each `subject_alternative_domain`
-  count = 1 + local.prom_records_count
-
-  name       = aws_acm_certificate.prometheus_cert.domain_validation_options[count.index]["resource_record_name"]
-  type       = aws_acm_certificate.prometheus_cert.domain_validation_options[count.index]["resource_record_type"]
-  zone_id    = var.zone_id
-  records    = [aws_acm_certificate.prometheus_cert.domain_validation_options[count.index]["resource_record_value"]]
-  ttl        = 60
+  for_each = {
```
blairboy362, line 109: I'm not entirely convinced this will match the `1 +` logic so I'd appreciate someone familiar with this to check this carefully.

philandstuff (dismissed):

terraform/modules/alertmanager/nlb.tf
philandstuff: I think this is fine as long as it works at all.  The `1 + length` logic is just an expression which should match the expected size of `aws_acm_certificate.alertmanager_cert.domain_validation_options` without having to dynamically depend on it, because such dynamic dependencies made terraform sad.

philandstuff: see https://www.terraform.io/docs/configuration/expressions.html#values-not-yet-known

> The count meta-argument for resources cannot be unknown, since it must be evaluated during the plan phase to determine how many instances are to be created.

blairboy362: :+1:

philandstuff (approved):

PR #410, opened 2020-08-06
Include changes to ECS ALB config in pipeline trigger

blairboy362:


szd55gds (approved):
LGTM

philandstuff:
Yeah those ALBs only front alertmanager, so they shouldn't trigger prometheus jobs...

philandstuff:
maybe the paths should be added to `alertmanager-git` instead?

philandstuff:
Argh sorry for triple-post. The problem is that the ALBs are managed in a whole separate project (`terraform/projects/app-ecs-albs-{staging,production}`) and that this project is not CD'd at the moment.  So we need to apply manually or set up a new concourse job.

blairboy362:
> Argh sorry for triple-post. The problem is that the ALBs are managed in a whole separate project (`terraform/projects/app-ecs-albs-{staging,production}`) and that this project is not CD'd at the moment. So we need to apply manually or set up a new concourse job.

Yes, I saw that. I'll tackle this when I've fixed the other stuff that's currently red.

PR #411, opened 2020-08-06
Fix task image build

blairboy362:
The terraform-provider-pass application no longer successfully `go gets`
in golang 1.10; it requires at least 1.11 (1.13 more desirable though.
Getting to golang > 1.10 requires a newer ubuntu.


szd55gds (approved):
LGTM

PR #416, opened 2020-08-13
revert alertmanager back to ALB

philandstuff:
This adds back an ALB for serving alertmanager traffic.

It can't go live because adding a target group to an ECS Service
forces destroy/recreate and we don't want to risk an alertmanager
outage.

### DIFF
--- a/terraform/modules/alertmanager/alertmanager-service.tf
+++ b/terraform/modules/alertmanager/alertmanager-service.tf
@@ -113,8 +113,11 @@ resource "aws_ecs_service" "alertmanager_nlb" {
 }
 
 resource "aws_ecs_service" "alertmanager_alb" {
-  count           = length(data.aws_subnet.private_subnets)
-  name            = "${var.environment}-alertmanager-alb-${data.aws_subnet.private_subnets[count.index].availability_zone}"
+  for_each = {
+    for _, subnet in data.aws_subnet.private_subnets :
+    subnet.id => subnet.availability_zone
+  }
+  name            = "${var.environment}-alertmanager-alb-${each.value}"
   cluster         = "${var.environment}-ecs-monitoring"
   task_definition = aws_ecs_task_definition.alertmanager_nlb.arn
   desired_count   = 1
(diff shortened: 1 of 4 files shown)


adityapahuja:
There is a commit which has WIP (Work In Progress).

philandstuff:
I mean yes, but future commits finished that work. Do you want me to rebase to remove "WIP" from the commit title?

adityapahuja:
We don't need to. At least we know that the commit is half done.

philandstuff:
@adityapahuja I've removed "WIP" from teh commit

terraform/modules/alertmanager/alertmanager-service.tf
```diff
@@ -112,6 +112,39 @@ resource "aws_ecs_service" "alertmanager_nlb" {
   }
 }
 
+resource "aws_ecs_service" "alertmanager_alb" {
+  for_each = {
+    for _, subnet in data.aws_subnet.private_subnets :
```
adityapahuja, line 117: What does `_` mean?

terraform/modules/alertmanager/alertmanager-service.tf
```diff
@@ -112,6 +112,39 @@ resource "aws_ecs_service" "alertmanager_nlb" {
...
+  launch_type     = "FARGATE"
+
+  load_balancer {
+    target_group_arn = aws_lb_target_group.alertmanager.arn
+    container_name   = "alertmanager"
+    container_port   = 9093
+  }
+
+  load_balancer {
+    target_group_arn = aws_lb_target_group.alertmanager_per_az[each.value].arn
```
chrisfarms, line 133: am I being blind or does this target group not exist?

terraform/modules/alertmanager/alertmanager-service.tf
philandstuff: It's a convention in some languages (maybe not terraform?) to mean "this is a variable I have to set but which I don't care about".  I only want the values from the map, not the keys.

terraform/modules/alertmanager/alertmanager-service.tf
philandstuff: it's in `alb.tf`

philandstuff: ...which is an untracked file. D'oh!

adityapahuja: I can confirm `alertmanager_per_az` does not exist.

philandstuff: now added

chrisfarms (approved):

PR #428, opened 2020-09-08
Trim trailing full stop to try to get AlertManager emails flowing again

Krenair:
These appear to have been broken since 23rd (ish) of January and the error AlertManager is outputting is this:
level=error ts=2020-09-08T08:11:47.804Z caller=dispatch.go:309 component=dispatcher msg="Notify for alerts failed" num_alerts=1 err="re-observe-ticket-alert/email[0]: notify retry canceled due to unrecoverable error after 1 attempts: parse 'from' addresses: mail: no angle-addr"
With the from address being "alerts@monitoring.gds-reliability.engineering."

It looks like Golang has never allowed the trailing full stop, so I'm suspicious this was caused by Terraform doing something related to https://github.com/terraform-providers/terraform-provider-aws/issues/241, maybe while applying 050dfc9513ca68a517c6582aecbdaea466a4864c

### DIFF
--- a/terraform/modules/alertmanager/alertmanager-service.tf
+++ b/terraform/modules/alertmanager/alertmanager-service.tf
@@ -179,6 +179,10 @@ data "pass_password" "verify_prod_cronitor" {
   path = "cronitor/verify-prod-url"
 }
 
+locals {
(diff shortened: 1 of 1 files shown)


szd55gds (approved):
LGTM

PR #430, opened 2020-09-23
observe alerts: add RE_Observe_No_Successful_Updates alert

risicle:
https://trello.com/c/vAmktonY

This is an alert we want to fire when we haven't seen any successful targets updates in a region for a certain period. updates are triggered through the /update-targets endpoint, so luckily we should just be able to look at the 200 responses to these requests to see successful updates.

Wrap in an outer `sum()` aggregation to leave open the possibility of multiple app instances.

Open to discussions of what the time window(s) and severity for this should be. I've initially chosen an `increase()` window of 30m, a `for:` period of 12h, and `ticket` severity.

The manual link doesn't exist (yet).

### DIFF
--- a/terraform/modules/prom-ec2/alerts-config/alerts/observe-alerts.yml
+++ b/terraform/modules/prom-ec2/alerts-config/alerts/observe-alerts.yml
@@ -95,6 +95,17 @@ groups:
         message: "One of the {{ $labels.job }} targets has been down for 24 hours"
         runbook: "https://re-team-manual.cloudapps.digital/prometheus-for-gds-paas-users.html#re-observe-target-down"
 
+  - alert: RE_Observe_No_Successful_Updates
+    expr: sum(increase(observe_broker_http_requests_total{code="200", path="/update-targets", method="post"}[30m])) by (region) == 0
+    for: 12h
+    labels:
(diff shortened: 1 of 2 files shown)


risicle:
I've added another commit on this which should (I _think_) annotate a `region="ireland"` label on to ireland's metrics. However I'm worried about this breaking anyones existing setups. It's the kind of thing that looks like it might not have been done for a reason, because it's so simple to copy the config from the `paas-london-targets` job. Having no label _does_ fit in with everything else in paas ireland being basically unmarked as such almost everywhere else. Thoughts?

chrisfarms:
> It's the kind of thing that looks like it might not have been done for a reason

I think Observe pre-dates PaaS having multiple regions (or at least pre-dates anyone in GDS using any region other than ireland), so I think it's most likely that the tags were only added to london when the deployment was added there.

It think it's worth adding the label.

chrisfarms (approved):
Looks good

I think 12hrs is fine to start ... and `ticket` seems like the right severity :+1:

risicle:
Oh dear smoke test failed on prom-3 going to staging.

risicle:
Re-ran & it passed :+1:

PR #433, opened 2020-10-12
Remove prometheus alerts for coronavirus services

gclssvglx:
What

The 3 coronavirus services (find support, business volunteer and
vulnerable people) are being shutdown and their infrastructure
removed.

Why

These services are either no longer required, will or have been
re-written or re-created, or have been migrated.

[Trello](https://trello.com/c/fcOEavQD/558-do-this-first-remove-prometheus-alertmanager-monitoring)


Krenair (changes requested):
You should also remove references from terraform/modules/prom-ec2/paas-config/main.tf and terraform/modules/alertmanager/templates/alertmanager.tpl

gclssvglx:
> You should also remove references from terraform/modules/prom-ec2/paas-config/main.tf and terraform/modules/alertmanager/templates/alertmanager.tpl

Thank you @Krenair 👍

Krenair (approved):
Sounds good to me but I'd like confirmation from another person involved in these services

gclssvglx:
> Sounds good to me but I'd like confirmation from another person involved in these services

Totally understand - thanks @Krenair

reggieb (approved):
LGTM

PR #437, opened 2020-11-11
Add an alert for when Notify disk space fills up

CrystalPea:
We want to know when disk space for our apps if filling up,
so that we can re-deploy those apps and avoid running out of
disk space.

In the future we may use this alert to automatically do the
re-deploy.

The threshold was chosen at 75%, because it takes one day for apps with quickly filling up disks to fill up 25% disk space. That means if the alert is triggered in the evening, after work, we still should have time in the morning to re-deploy the app and reset the disk usage.
It should not be much lower though, because then we can start getting too many alerts.

QUESTIONS:
- Can I test it at this point, before we merge it with master? Or do we do that afterwards?
- do we set up a receiver before or after? (best method would be a Zendesk ticket, because it's an alert that needs to be followed by a concrete action. Open ticket will remind person on support to solve it.)

### DIFF
--- a/terraform/modules/alertmanager/alertmanager-service.tf
+++ b/terraform/modules/alertmanager/alertmanager-service.tf
@@ -155,6 +155,10 @@ data "pass_password" "registers_zendesk" {
   path = "receivers/registers/zendesk"
 }
 
+data "pass_password" "notify_zendesk" {
+  path = "receivers/notify/zendesk"
+}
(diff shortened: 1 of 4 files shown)


Krenair:
(answers in Slack)

risicle (approved):

Krenair (approved):

PR #442, opened 2020-11-16
fix terraform 0.13

philandstuff:
This failed with:

    There are some problems with the configuration, described below.

    The Terraform configuration must be valid before initialization so that
    Terraform can determine which modules and providers need to be installed.

    Error: Invalid version constraint

    on versions.tf line 3, in terraform:
    3: aws = {
    4: source = "hashicorp/aws"
    5: }

    A string value is required for aws.

I'm not sure what's going on here but I *think* that a `version` is
required at the top-level module (but not at modules included from
elsewhere?).

### DIFF
--- a/terraform/projects/alertmanager-production/main.tf
+++ b/terraform/projects/alertmanager-production/main.tf
@@ -31,12 +31,7 @@ terraform {
 }
 
 provider "aws" {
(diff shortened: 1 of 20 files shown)


szd55gds (approved):
This looks ok to me

PR #444, opened 2020-11-16
Update SHA256 for terraform download

philandstuff:
This Dockerfile kept the terraform version quite far away from the
validating hash.  I've pulled the hash to live next to the version so
that people might actually know/remember to update both.


chrisfarms (approved):
heh yes I would totally have missed that too!

PR #446, opened 2020-11-17
Add link to grafana for notify disk space alerts

idavidmcdonald:
Adds an easy link to go look at grafana

Also move the summary and message into a single annotation as there is
no readability benefit in the zendesk ticket of having them as separate
things


cbaines (approved):
Looks good to me :+1:

CrystalPea (approved):

PR #447, opened 2020-11-18
Add notify alert for disk space 95% full

CrystalPea:
This alert should page Notify's pagerduty, as it means disk space for a production app will run out shortly.

We also have a Zendesk ticket level alert for disk space being
75% full should. In most cases we should be able to react to that
less urgent alert in time. This will prevent this pagerduty fallback
alert from being set off.

Pagerduty alert is in case disk space fills up before we see the
Zendesk ticket.

Note, we see invididual commit messages for why we decided this is a p2 pagerduty rather than a p1.

Will require our pagerduty p2 service key being added to the autom8 password store for this to be merged.

### DIFF
--- a/terraform/modules/alertmanager/alertmanager-service.tf
+++ b/terraform/modules/alertmanager/alertmanager-service.tf
@@ -161,6 +161,10 @@ data "pass_password" "notify_zendesk" {
   path = "receivers/notify/zendesk"
 }
 
+data "pass_password" "notify_p2_pagerduty_key" {
+  path = "receivers/notify/p2_pagerduty"
+}
+
 data "pass_password" "autom8_email" {
   path = "receivers/autom8/email"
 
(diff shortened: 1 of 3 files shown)


cbaines (dismissed):
Seems fine to me :+1:

terraform/modules/alertmanager/alertmanager-service.tf
```diff
@@ -161,6 +161,10 @@ data "pass_password" "notify_zendesk" {
   path = "receivers/notify/zendesk"
 }
 
+data "pass_password" "notify_p2_pagerduty_key" {
```
cbaines, line 164: On the naming of the key, is the key specific to P2 incidents, or is it not specific to the priority?

cbaines:
This generally looks fine to me :+1: I made one naming related comment.

> Will require our pagerduty p2 service key being added to the autom8 password store for this to be merged.

Is this something you'd like my help with, or is this something you're doing?

terraform/modules/alertmanager/alertmanager-service.tf
idavidmcdonald: It is specific to p2 incidents. We have one key for p1 incidents and one key for p2 incidents (although this PR only adds the p2 key as we don't yet require the p1 key in any of these alerts)

idavidmcdonald:
We do not have access to the autom8 password store so we will need to send you the pagerduty p2 service key for you to add please

terraform/modules/alertmanager/alertmanager-service.tf
cbaines: Cool, good to know :+1:

cbaines (approved):

PR #450, opened 2020-11-19
Example of custom alertmanager email template

idavidmcdonald:
This is what it would look like to do custom email templates to zendesk. It would mean zendesk alerts could look much prettier and only contain labels that teams need if they want to define their own templates.

Note, I haven't tested it and I haven't checked to see what the email would actually end up looking like but could put some time into this if need be.

Note, for zendesk, I believe all the html gets stripped and all we are left with is the the text so the approach would be to go for almost purely text based alerts rather than html.

Currently alerts look like:
image

For further reading/context

https://prometheus.io/docs/alerting/latest/configuration/#email_config
https://prometheus.io/blog/2016/03/03/custom-alertmanager-templates/
https://github.com/prometheus/alertmanager/blob/master/template/email.html

Problem that would need solving is how to get the custom templates files on to the alertmanager instance. At the moment it looks like this is being done using the ECS task entrypoint (https://github.com/alphagov/prometheus-aws-configuration-beta/blob/master/terraform/modules/alertmanager/task-definitions/alertmanager.json#L30) but might want to go for the approach being done for prometheus of putting the files in an s3 bucket

### DIFF
--- a/terraform/modules/alertmanager/templates/alertmanager.tpl
+++ b/terraform/modules/alertmanager/templates/alertmanager.tpl
@@ -128,7 +128,7 @@ receivers:
 - name: "notify-tickets"
   email_configs:
   - to: "${notify_zendesk}"
-  - html: '{{ template "zendesk.notify.html" . }}'
+  - html: '{{ template "zendesk.notify.text" . }}'
 - name: "observe-cronitor"
   webhook_configs:
   - send_resolved: false
--- a/terraform/modules/alertmanager/templates/alertmanager.tpl
+++ b/terraform/modules/alertmanager/templates/alertmanager.tpl
@@ -9,6 +9,7 @@ global:
 
 templates:
 - '/etc/alertmanager/default.tmpl'
+- '/etc/alertmanager/notify.tmpl'
 
 route:
   receiver: "re-observe-pagerduty"
@@ -127,6 +128,7 @@ receivers:
 - name: "notify-tickets"
   email_configs:
   - to: "${notify_zendesk}"
+  - html: '{{ template "zendesk.notify.html" . }}'
 - name: "observe-cronitor"
   webhook_configs:
   - send_resolved: false
--- a/terraform/modules/alertmanager/templates/notify.tmpl
+++ b/terraform/modules/alertmanager/templates/notify.tmpl
@@ -1,4 +1,4 @@
-{{ define "zendesk.notify.html" }}
+{{ define "zendesk.notify.text" }}
 
 {{ .Alerts | len }} alert{{ if gt (len .Alerts) 1 }}s{{ end }} for {{ range .GroupLabels.SortedPairs }}
 {{ .Name }}={{ .Value }} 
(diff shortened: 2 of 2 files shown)


terraform/modules/alertmanager/templates/alertmanager.tpl
```diff
@@ -9,6 +9,7 @@ global:
 
 templates:
 - '/etc/alertmanager/default.tmpl'
+- '/etc/alertmanager/notify.tmpl'
```
philandstuff, line 12: we'll either need to add this into `data.template_file.alertmanager_nlb_container_defn` next to `default.tmpl`, or (probably easier for now) whack it into `default.tmpl`.

terraform/modules/alertmanager/templates/alertmanager.tpl
```diff
@@ -127,6 +128,7 @@ receivers:
 - name: "notify-tickets"
   email_configs:
   - to: "${notify_zendesk}"
+  - html: '{{ template "zendesk.notify.html" . }}'
```
philandstuff, line 131:
```suggestion
  - html: '{{ template "zendesk.notify.text" . }}'
```
it's not html, as you say in the PR description :)

terraform/modules/alertmanager/templates/notify.tmpl
```diff
@@ -0,0 +1,18 @@
+{{ define "zendesk.notify.html" }}
```
philandstuff, line 1:
```suggestion
{{ define "zendesk.notify.text" }}
```

philandstuff:
This looks great! I'd say let's fix it up and merge it relatively quickly, then iterate.  Other feedback:

- I'd like us to use it for RE zendesk too (while we still use zendesk), unless there's anything terribly notify-specific you want to do?
- It'd be nice to explicitly call out certain annotations (`runbook` if present, `summary`/`details`; this is where standardised annotations really shine)

idavidmcdonald:
Happy for this to become the default zendesk template. I can't immediately think of a reason to start with this as Notify specific. If we can find one later then it can be changed.

A few things to note
- I haven't tested it so definitely needs a test to see if the template will render correctly
- I haven't really thought about would go in the email
- At the moment, the template I threw together only works for firing alerts, it doesn't work for resolved alerts (see the default for one that does https://github.com/prometheus/alertmanager/blob/master/template/email.html). So would only work for teams that don't use `send_resolved` as `True`.


My initial thoughts about what should/shouldn't go in the alert
- I don't ever feel a need to see the alert in alertmanager so I probably wouldn't include a link
- I don't see a particular need to show the label for product or severity, it's quite obvious that it's a ticket alert as it's a ticket and it's quite obvious it's about Notify because the alertname has it in and this is in the notify zendesk queue
- I do like having the links at the very bottom saying it's come from prometheus and alertmanager, just so developers who don't see this often know what the source of the alert is rather than it being opaque because they never touch the monitoring infrastructure
- we may also want to think about the subject too

Note, some of those points/thoughts may not work for all teams

For my example alert I shared in the screenshot above, for me the ideal alert would be something like


```
Alert for GOVUK_Notify_Disk_75_percent_full

Action required:
Test_disk_space_alert needs to be re-deployed to avoid running out of disk space

Details:
app: test_disk_space_alert
space: production

Dashboard:
link to grafana here

Sent by AlertManager (https://alerts.monitoring.gds-reliability.engineering)

```

idavidmcdonald:
Going to close this as just doing a bit of tidy up. I don't think we will get round to any improvements on this.

PR #452, opened 2020-11-19
Scrape new dcs prom

philandstuff:
[card]: https://trello.com/c/wgYRVMKq

### DIFF
--- a/terraform/projects/prom-ec2/paas-production/extra-prometheus-scrape-configs.yml.tpl
+++ b/terraform/projects/prom-ec2/paas-production/extra-prometheus-scrape-configs.yml.tpl
@@ -1,3 +1,22 @@
+- job_name: dcs-federate
+  scheme: https
+  honor_labels: true
+  honor_timestamps: true
+  metrics_path: '/federate'
+  params:
+    "match[]":
+    # fetch everything (via https://stackoverflow.com/a/39253848 )
+    - '{__name__=~".+"}'
+  static_configs:
+  - targets:
+    - dcs-build-internal-prometheus.london.cloudapps.digital
+    labels:
+      federated-from: dcs-build-internal-prometheus.london.cloudapps.digital
+  - targets:
+    - dcs-integration-internal-prometheus.london.cloudapps.digital
+    labels:
+      federated-from: dcs-integration-internal-prometheus.london.cloudapps.digital
+
 - job_name: paas_elasticsearch_for_dm
   scheme: https
   basic_auth:


terraform/projects/prom-ec2/paas-production/extra-prometheus-scrape-configs.yml.tpl
```diff
@@ -1,3 +1,19 @@
...
+  honor_labels: true
+  honor_timestamps: true
+  metrics_path: '/federate'
+  params:
+    "match[]":
+    # fetch everything (via https://stackoverflow.com/a/39253848 )
+    - '{__name__=~".+"}'
+  static_configs:
+  - targets:
+    - dcs-sandbox-prometheus.cloudapps.digital
```
philandstuff, line 12: i made this domain up, i don't know what the correct thing is

terraform/projects/prom-ec2/paas-production/extra-prometheus-scrape-configs.yml.tpl
```diff
@@ -1,3 +1,27 @@
...
+    labels: # although maybe these could be external_labels on the
+            # scraped prometheus?
+      space: sandbox
+      org: gds-document-checking-service
+      region: ireland
+  - targets:
+    - dcs-integration-prometheus.cloudapps.digital
```
philandstuff, line 19: this is also made up

terraform/projects/prom-ec2/paas-production/extra-prometheus-scrape-configs.yml.tpl
```diff
@@ -1,3 +1,27 @@
+- job_name: dcs-federate
+  scheme: https
+  honor_labels: true
+  honor_timestamps: true
+  metrics_path: '/federate'
```
risicle, line 5: Oh. How long were you going to wait for me to realize I'd mapped the wrong route path on the internal prometheus (I thought it used `/metrics` like anything else)

terraform/projects/prom-ec2/paas-production/extra-prometheus-scrape-configs.yml.tpl
```diff
@@ -1,3 +1,27 @@
...
+  params:
+    "match[]":
+    # fetch everything (via https://stackoverflow.com/a/39253848 )
+    - '{__name__=~".+"}'
+  static_configs:
+  - targets:
+    - dcs-sandbox-prometheus.cloudapps.digital
+    labels: # although maybe these could be external_labels on the
+            # scraped prometheus?
```
risicle, line 14: I think we should only annotate the hostname that the metrics were federated _from_ here, because that's all we know for sure at this end. I think it would be better to annotate space/org/region/app name from the federated end, where it can be injected via the terraform template and is less likely to get out of sync.

terraform/projects/prom-ec2/paas-production/extra-prometheus-scrape-configs.yml.tpl
risicle: All domains are made up, man.

risicle:
I have amended the config here to only add the hostname of the internal prometheus, hoping we can add more labels from the other end more reliably.

szd55gds (dismissed):
LGTM

cbaines (approved):

PR #460, opened 2020-12-10
Fix typo (quartile vs. percentile)

richardTowers:
Quartile refers specifically to the four quantiles that split a dataset into quarters.
It doesn't make sense to talk about a 90th quartile - you can only have 1st, 2nd and 3rd quartiles.

Percentile is more accurate in this case. I think it's also more widely
used / commonly understood than "quartile", so maybe a bit easier to
understand.

Unlikely that this would confuse anyone in practice, but if it's going
to end up in an alert we should make it as clear as possible.


bevanloon (dismissed):

cbaines (dismissed):
I'd normally say percentile, but this is fine as well.

richardTowers:
👍  I agree, percentile is better. I also spotted the same thing in the alert name, so I've force pushed an update. Sorry to make you review it again 😓

bevanloon (approved):

Krenair (approved):

szd55gds (approved):
LGTM

PR #461, opened 2020-12-15
Add initial configuration for DCS on the PaaS

cbaines:
This commit adds the Prometheus alerts configuration, as well as some
minimal alertmanager config which should avoid any premature pages for
people.

Once the alerts are in place, I'll setup some proper routing.

### DIFF
--- a/terraform/modules/alertmanager/templates/alertmanager.tpl
+++ b/terraform/modules/alertmanager/templates/alertmanager.tpl
@@ -56,6 +56,9 @@ route:
     match:
       layer: "infra"
       severity: "ticket"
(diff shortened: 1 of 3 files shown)


risicle:
I'm curious - how did you develop/test this?

cbaines:
> I'm curious - how did you develop/test this?

It's an adaptation of the configuration here https://github.com/alphagov/doc-checking/blob/master/configuration/dcs-helm/templates/prometheus-alerts.yaml

I adapted and tested the queries by just copying them in to the Prometheus web interface. As for the YAML, :crossed_fingers:

risicle (approved):

PR #463, opened 2021-01-05
Do not trigger accounts 5xx alert from the error app

barrucadu:
This is an nginx-buildpack app put up if we need to take the normal accounts system down, and always returns a 503 error.

### DIFF
--- a/terraform/modules/prom-ec2/alerts-config/alerts/govuk-accounts-alerts.yml
+++ b/terraform/modules/prom-ec2/alerts-config/alerts/govuk-accounts-alerts.yml
@@ -2,7 +2,7 @@ groups:
 - name: GovukAccounts
   rules:
   - alert: GovukAccounts_AppRequestsExcess5xx
-    expr: sum without(exported_instance, status_range) (rate(requests{org="govuk-accounts", space="production", status_range="5xx"}[5m])) / sum without(exported_instance, status_range) (rate(requests{org="govuk-accounts", space="production"}[5m])) >= 0.25
+    expr: sum without(exported_instance, status_range) (rate(requests{org="govuk-accounts", space="production", status_range="5xx", app!="govuk-account-static-errors"}[5m])) / sum without(exported_instance, status_range) (rate(requests{org="govuk-accounts", space="production"}[5m])) >= 0.25
     for: 120s
     labels:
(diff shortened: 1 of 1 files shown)


terraform/modules/prom-ec2/alerts-config/alerts/govuk-accounts-alerts.yml
```diff
@@ -2,7 +2,7 @@ groups:
 - name: GovukAccounts
   rules:
   - alert: GovukAccounts_AppRequestsExcess5xx
-    expr: sum without(exported_instance, status_range) (rate(requests{org="govuk-accounts", space="production", status_range="5xx"}[5m])) / sum without(exported_instance, status_range) (rate(requests{org="govuk-accounts", space="production"}[5m])) >= 0.25
+    expr: sum without(exported_instance, status_range) (rate(requests{org="govuk-accounts", space="production", status_range="5xx", app!="govuk-account-static-errors"}[5m])) / sum without(exported_instance, status_range) (rate(requests{org="govuk-accounts", space="production"}[5m])) >= 0.25
```
brucebolt, line 5: Does the denominator of the fraction also need an exclusion for the static errors app?  Otherwise the percent won't be quite right.

barrucadu: No, Prometheus matches labels when doing arithmetic on metrics.  See https://prometheus.io/docs/prometheus/latest/querying/operators/#vector-matching

cbaines (approved):
Looks good to me :+1:

terraform/modules/prom-ec2/alerts-config/alerts/govuk-accounts-alerts.yml
huwd: :exploding_head:

brucebolt (approved):

barrucadu:
@cbaines Would you be able to merge and deploy this for me?

cbaines:
> @cbaines Would you be able to merge and deploy this for me?

It's all automated, but I've checked Prometheus has the updated config, and I believe it does.

barrucadu:
Thanks

PR #467, opened 2021-02-24
add /last-config endpoint to prometheus nodes 

risicle:
https://trello.com/c/H5pjav8d

This should allow us to reliably tell when a new config has successfully been applied to the running prometheus, whether the machine needed to be restarted or not.

Monitoring `prometheus_config_last_reload_successful` and/or `prometheus_config_last_reload_success_timestamp_seconds` doesn't appear to be sufficient in our case as:

 - bad configs don't actually tend to get _seen_ by the running prometheus daemon as they get rejected early by the init script
 - they would always get tripped on a restart-requiring deployment, whether the desired config has made it out to the node or not.

Wrap the updates in a `flock` call to ensure we don't have a funny race condition involving the reading of the config file and I can cast the thought from my brain entirely. 30s acquisition timeout when used from the `s3 sync` end because we know another attempt will be along in a minute to try again. No timeout on the reload-loop end as we'll only get one notification per modification and can't afford to throw that away.

Considered making this a prometheus-metrics-formatted file so we could monitor it from prometheus, but that's harder to consume in a script (no `jq` equivalent). We could always spit out both formats if we decided we wanted that down the line.

### DIFF
--- a/terraform/modules/prom-ec2/paas-config/outputs.tf
+++ b/terraform/modules/prom-ec2/paas-config/outputs.tf
@@ -1,0 +1,3 @@
+output "prometheus_config_etag" {
+  value = aws_s3_bucket_object.prometheus_config.etag
+}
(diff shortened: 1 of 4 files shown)


Krenair (approved):

risicle:
Bugger I meant to add the correct mimetype for this

PR #469, opened 2021-02-26
ci: fix deploy pipeline's healthchecks to check hash of prometheus' last config

risicle:
https://trello.com/c/H5pjav8d

This is mostly a matter of hooking stuff together.

Firstly I had to add concourse's IPs to prometheus' allow-list to be able to poke the `/last-config` endpoint. Previously we were able to do without this as the `/healthz` endpoint was not protected by nginx's auth restriction.

I made the `http-ping` task capable of accepting an optional `jq` script for testing the response body against, and generate this in a little intermediate task using the etag output from the terraform run.

Tested with a no-change re-deployment, an S3-config-only change, and with a full instance-re-provision deployment. Works.


ci/tasks/generate-prometheus-test-jq.yml
```diff
@@ -0,0 +1,11 @@
...
+inputs:
+  - name: input
+outputs:
+  - name: output
+run:
+  path: sh
+  args:
+    - -euxc
+    - |
+      echo ".last_successful_config == $(jq '.prometheus_config_etag.value' input/terraform-outputs.json)" > output/test.jq
```
cbaines, line 11: I guess `jq` default quoting for outputted strings makes the `$(...)` here work, which is interesting...

cbaines (approved):
Looks OK to me :+1:

ci/tasks/generate-prometheus-test-jq.yml
risicle: yes i thought it was quite neat

PR #472, opened 2021-03-05
ci smoke tests: rename previous tests "config tests", add actual smoke tests of prometheus

risicle:
https://trello.com/c/H5pjav8d

Previous tests only tested nginx and the config replacement system was happy, we should also check prometheus is happily running before we continue.


cbaines (approved):
The first line of your commit message is too long, but apart from that, this looks reasonable.

risicle:
I know, but it was either leave off the categorization of the change or leave off one of the key things it does. And terser language just made it nonsense.

PR #473, opened 2021-03-09
ci: consolidate prometheus-aws-configuration-beta git resources

risicle:
Before:
Screenshot 2021-03-09 at 12 09 11

After:
Screenshot 2021-03-09 at 11 25 53

(yes, tiny text, sorry)

Going from 4 to 2. Most jobs are fairly fast when applying an empty change and having many split git resources creates a bunch of weird race conditions. One task that is expensive for every run is building the task image, so move that to its own job (from the `configure-pipeline` job) with its own dedicated git resource. This way simultaneous changes to the task image and the rest of the repo should be the only case that can introduce weird race conditions.

### DIFF
--- a/ci/deploy.vars.default.yml
+++ b/ci/deploy.vars.default.yml
@@ -1,0 +1,2 @@
+background_image: ""
+prometheus-aws-configuration-beta-branch: master
--- a/ci/deploy.yml
+++ b/ci/deploy.yml
@@ -1,3 +1,6 @@
+display:
+  background_image: ((background-image))
+
 resource_types:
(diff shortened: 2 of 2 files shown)


risicle:
I note that this change does indeed make having separate `-common`, `-prometheus`, and `-alertmanager` jobs rather pointless, but TBH I think I'm of the opinion that they are kinda pointless, and a bit of a liability when it comes to possible failure scenarios and different versions of the resources bubbling through the pipeline in very close succession.

Krenair (approved):

PR #476, opened 2021-04-06
BAU: remove references to GSP in alertmanager config

dbes-gds:
## What

Remove references to GSP in alertmanager config.

## Why

So that the verify-dcs-gsp-alerts can be renamed to di-dcs-2ndline now that DCS has moved to PaaS.


ghost (approved):
👍 . It looks like all the Autom8 alerts here are just for GSP but it's just a name change so i think that is fine.

chrisclayson (approved):

Krenair (approved):

PR #480, opened 2021-07-12
Rescue slack-configs YAML anchor removed in #479

Krenair:


risicle (approved):
assume this is the only remaining place it's used

stephengrier (approved):

PR #483, opened 2021-07-23
[#178898815] alertmanager deployment: wait for ecs service to be stable before proceeding

risicle:
The hope is that a broken deployment will be caught even if the current release doesn't cause an ECS deployment.

This does seem to have the desired effect in practise when the deployment is deliberately broken.


risicle:
Sidenote: working on this reminded me how much I'd like to get #473 merged because its annoying and nerve-wracking having every push you make potentially triggering multiple builds in unexpected orders.

tommrr (approved):

PR #485, opened 2021-07-28
[#178898838] terraform: ensure alertmanager configs are valid yaml/json

risicle:
Having to use an external module here because terraform has no native support for throwing errors deliberately.

Do take a look at it and see if you think it's weird: https://github.com/Invicton-Labs/terraform-null-assertion

See how it makes a deliberate call to a non-existent program as a means of throwing an error during the plan phase. Liability?

Screenshot 2021-07-28 at 13 38 11 And as a result its error message looks a bit weird.

The other way of doing it was to use a special provider, but that was even weirder and probably more of a liability.

Using `depends_on` ensures even a targeted/partial terraform apply should require this check to be performed.

If we're ok with this, do you think I should also put similar checks on the alert yamls in `terraform/modules/prom-ec2/paas-config/`?

### DIFF
--- a/terraform/modules/alertmanager/alertmanager-service.tf
+++ b/terraform/modules/alertmanager/alertmanager-service.tf
@@ -72,6 +72,18 @@ data "template_file" "alertmanager_nlb_container_defn" {
     log_group                  = aws_cloudwatch_log_group.task_logs.name
     region                     = var.aws_region
   }
+
+  depends_on = [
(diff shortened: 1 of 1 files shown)


tommrr (dismissed):

risicle:
Failure now using updated external module 😆 

Screenshot 2021-08-02 at 12 48 01

Certainly uglier but doesn't do the weird `external` call-out.

tommrr (approved):

PR #487, opened 2021-10-28
Delete GOV.UK accounts alerts

barrucadu:
We're retiring these apps now.


risicle (approved):
Cool - let's throw it at staging.

PR #488, opened 2023-02-01
BAU - Add latest IP addresses to alertmanager

ghost:
- The list did not included the latest GDS IP addresses


elena-vi (approved):

risicle:
This broke the deployment for alertmanager, I had to disable this version in the concourse pipeline and run it again (which appears to work now)

Error was:

```
Error: Error authorizing security group rule type ingress: InvalidParameterValue: The same permission must not appear multiple times
```

PR #489, opened 2023-04-05
BAU - Remove duplicated IP addresses

ghost:
- The pipeline broke as there multiple entries with the same IP address


owenblacker (approved):
LGTM

